PR #5, opened 2026-04-16
perf: use getSession() instead of getUser() in client components

ArnasDon:
## Summary

Cuts ~150-400ms off every page load by replacing \`supabase.auth.getUser()\` (network round-trip) with \`supabase.auth.getSession()\` (reads from localStorage, no network) in client components.

## Why this matters

Every client component that needed the current user was doing:

\`\`\`
Page mount
  -> supabase.auth.getUser()  // network call: 150-400ms
  -> query data                // network call: 150-400ms
  -> render
\`\`\`

Two serial round-trips before the page shows anything. Your Supabase project is in EU (Paris) — if you're on a US-hosted deployment or hitting it from the US, that's ~200-300ms per round-trip. Cumulative: 500-800ms of latency per page.

\`getSession()\` reads from localStorage synchronously (within the SDK) without contacting Supabase. One round-trip saved per page.

## Changes

Replaced the pattern in 11 client files (10 components + 1 page):

- \`src/app/(dashboard)/inbox/page.tsx\`
- \`src/app/(dashboard)/pipelines/page.tsx\`
- \`src/components/contacts/*.tsx\` (3 files)
- \`src/components/inbox/contact-sidebar.tsx\`
- \`src/components/pipelines/deal-form.tsx\`, \`pipeline-settings.tsx\`
- \`src/components/settings/*.tsx\` (3 files)

## Intentionally NOT changed

- \`src/middleware.ts\` — runs server-side, needs JWT validation against Supabase
- \`src/app/api/whatsapp/*/route.ts\` — security-critical server code
- \`src/hooks/use-auth.ts\` — PR #4 handles this file separately

## Security note

\`getSession()\` trusts the locally-stored session. That's fine here because:
- Client components only use the user ID to filter RLS-protected queries
- The actual authorization boundary is **Row Level Security on the database** (e.g. \`auth.uid() = user_id\`)
- A forged user ID in the browser can't bypass server-side RLS policies
- API routes and middleware still use \`getUser()\` (verified against Supabase) for any real authorization decisions

## Test plan

- [ ] \`npm run dev\` locally → pages load visibly faster
- [ ] Settings page loads without hanging
- [ ] Contact form / deal form / template form / tag form submit correctly
- [ ] User sees their own data (RLS still enforced)

### DIFF
--- a/src/app/(dashboard)/inbox/page.tsx
+++ b/src/app/(dashboard)/inbox/page.tsx
@@ -25,8 +25,9 @@ export default function InboxPage() {
     const checkConnection = async () => {
       const supabase = createClient();
       const {
-        data: { user },
-      } = await supabase.auth.getUser();
+        data: { session },
+      } = await supabase.auth.getSession();
+      const user = session?.user;
 
       if (!user) return;
 
--- a/src/app/(dashboard)/pipelines/page.tsx
+++ b/src/app/(dashboard)/pipelines/page.tsx
@@ -85,8 +85,9 @@ export default function PipelinesPage() {
     setCreating(true);
 
     const {
-      data: { user },
-    } = await supabase.auth.getUser();
+      data: { session },
+    } = await supabase.auth.getSession();
+    const user = session?.user;
     if (!user) {
       setCreating(false);
       return;
--- a/src/components/contacts/contact-detail-view.tsx
+++ b/src/components/contacts/contact-detail-view.tsx
@@ -223,8 +223,9 @@ export function ContactDetailView({
     setSavingNote(true);
 
     const {
-      data: { user },
-    } = await supabase.auth.getUser();
+      data: { session },
+    } = await supabase.auth.getSession();
+    const user = session?.user;
     if (!user) {
       toast.error('Not authenticated');
       setSavingNote(false);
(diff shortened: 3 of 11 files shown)


giladamru:
## Second live verification pass

Second live pass (2026-09-20, fresh isolated stack, real Meta calls with the dev test number, no real recipient): an inbound PNG and an inbound "document" declared `text/html`, delivered as signed webhooks pointing at media ids uploaded to Meta, were fetched from Meta and mirrored into the private bucket with `media_bucket`/`media_path` set; headless Chromium logged in as the agent renders the image through `/api/media` → signed URL and the HTML document only ever arrives as a download named `probe.html` (page title never changes, no popup). `POST /api/whatsapp/send` with `media_path` downloaded the object under RLS, uploaded it to Meta by id and got a wamid back. Four `send_webhook` automations aimed at `127.0.0.1`, `127.0.0.1.nip.io`, `[::ffff:7f00:1]` and `169.254.169.254` all failed with `destination not allowed` and a listener on the target port saw zero requests. Also checked: `touch_presence` 204 as authenticated / 401 as anon, `peek_invitation` and `redeem_invitation` still callable, `merge_duplicate_contacts` and `_bcast_bump` 403 for an owner JWT; agent `POST /api/whatsapp/config` 403; broadcast with one unknown number 400 listing it; media proxy 120 × 404 then 429; invitation peek trips the per-token bucket after 30 even with a rotating `X-Forwarded-For`; v1 cursors with a V8-only date or a PostgREST filter fragment fall back to page one instead of erroring; v1 messages return signed Storage URLs.

PR #24, opened 2026-04-17
fix(broadcasts): set user_id on send, surface errors with a toast

ArnasDon:
## Summary

Fixes \"Send Broadcast does nothing\" — the wizard completed all 4 steps, the confirmation modal closed, and the flow silently no-op'd.

## Root cause

\`src/hooks/use-broadcast-sending.ts\`'s INSERT into \`broadcasts\` never set \`user_id\`. But \`broadcasts.user_id\` is \`NOT NULL\` and the RLS policy requires \`auth.uid() = user_id\` — so PostgREST rejected the INSERT (error 23502 / 42501). The hook threw, \`handleSend\` caught it with \`console.error\` only, and the UI had no idea anything happened. \`isProcessing\` flipped on then off in the same tick so the progress bar never even flashed.

The save-as-draft path (added in PR B) already sets \`user_id\` correctly — this is why draft works but send doesn't.

## Fix

- Resolve `session.user` at the start of \`createAndSendBroadcast\` and include \`user_id: user.id\` on the broadcasts INSERT.
- \`new/page.tsx handleSend\` now shows the error via a \`sonner\` toast so future failures surface immediately instead of vanishing.

## Test plan
- [ ] Send a broadcast to ≥1 contact → wizard redirects to \`/broadcasts/:id\` and the row shows up in the list
- [ ] Sign out → attempt send (edge case) → toast says \"You are not signed in.\"
- [ ] Force a Meta API error (invalid template_name) → hook still completes, individual recipients marked failed, broadcast row created (status=failed)

### DIFF
--- a/src/app/(dashboard)/broadcasts/new/page.tsx
+++ b/src/app/(dashboard)/broadcasts/new/page.tsx
@@ -2,6 +2,7 @@
 
 import { useState } from 'react';
 import { useRouter } from 'next/navigation';
+import { toast } from 'sonner';
 import { MessageTemplate } from '@/types';
 import { Step1ChooseTemplate } from '@/components/broadcasts/step1-choose-template';
 import { Step2SelectAudience } from '@/components/broadcasts/step2-select-audience';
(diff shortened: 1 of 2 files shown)


ArnasDon:
Superseded by #25 — that PR includes this user_id fix on top of the full B+C diff and promotes everything to main in one merge.

PR #52, opened 2026-04-19
fix(mobile + seo): broadcasts table responsive + opengraph-image render

ArnasDon:
## Summary

Last of the mobile-friendly series (following #48/#49/#50/#51), plus a latent production bug in the OG image endpoint that was throwing 500s on every request.

### Broadcasts table — responsive column tiers

The table rendered every column at every width, so a 375px phone had to swipe sideways just to read a name. Now it matches the tiering the contacts table already uses:

| Column | `<sm` | `sm+` | `md+` | `lg+` |
|---|---|---|---|---|
| Name | ✓ | ✓ | ✓ | ✓ |
| Status | ✓ | ✓ | ✓ | ✓ |
| Recipients | — | ✓ | ✓ | ✓ |
| Date | — | ✓ | ✓ | ✓ |
| Template | — | — | ✓ | ✓ |
| Delivery | — | — | — | ✓ |
| Read | — | — | — | ✓ |

### OG image endpoint — fixed 500 crashes

`/opengraph-image` was throwing on every request with two errors that Satori (the renderer behind `next/og`) is strict about:

1. `Expected <div> to have explicit "display: flex" if it has more than one child node.`
2. `Failed to download dynamic font for ✓` — the default font Satori loads doesn't cover the ✓ glyph.

Result: WhatsApp / LinkedIn / Slack shares of wacrm.tech were getting no link preview.

Fix:
- Every `<div>` with more than one child now has explicit `display: 'flex'`.
- The headline's inline `<span>` is now two `<span>`s inside a `display: flex; flex-wrap: wrap` parent.
- ✓ glyphs in the feature list → plain text dot separators.

Verified: `curl -sI http://localhost:3000/opengraph-image` → `HTTP/1.1 200 OK`, `content-type: image/png`.

## Test plan

- [ ] Broadcasts page on mobile (375px): only Name + Status columns visible, no horizontal scroll.
- [ ] Broadcasts page on a tablet (~768px): Name + Template + Recipients + Date + Status.
- [ ] Broadcasts page on desktop (≥1024px): all seven columns as before.
- [ ] `GET /opengraph-image` returns 200 `image/png` (no 500s in server logs).
- [ ] Paste the production URL into WhatsApp / LinkedIn / Slack after deploy → branded 1200×630 preview appears.

### DIFF
--- a/src/app/(dashboard)/broadcasts/page.tsx
+++ b/src/app/(dashboard)/broadcasts/page.tsx
@@ -211,14 +211,14 @@ export default function BroadcastsPage() {
             <TableHeader>
               <TableRow className="border-slate-800 hover:bg-transparent">
                 <TableHead className="text-slate-400">Name</TableHead>
-                <TableHead className="text-slate-400">Template</TableHead>
-                <TableHead className="text-slate-400 text-right">
+                <TableHead className="hidden text-slate-400 md:table-cell">Template</TableHead>
+                <TableHead className="hidden text-right text-slate-400 sm:table-cell">
                   Recipients
                 </TableHead>
-                <TableHead className="text-slate-400">Delivery</TableHead>
-                <TableHead className="text-slate-400">Read</TableHead>
+                <TableHead className="hidden text-slate-400 lg:table-cell">Delivery</TableHead>
+                <TableHead className="hidden text-slate-400 lg:table-cell">Read</TableHead>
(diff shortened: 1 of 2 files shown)


ArnasDon:
Superseded: the OG-image fix is already on main via PR #55; the broadcasts-table change is now in a dedicated PR.

PR #163, opened 2026-05-30
fix flow drag responsiveness and stale chunks

dhruv465:
## Summary
- let Next own /_next/static cache headers so dev chunks revalidate instead of serving stale module factories
- keep the flow canvas nodes controlled during drag with React Flow node changes
- persist generated dagre positions once so dragging a node does not make other nodes jump back to the origin

## Root cause
Custom cache headers forced /_next/static chunks to be immutable, which can leave Turbopack dev pages loading stale client modules. The flow canvas also rendered controlled nodes without onNodesChange, and dagre positions were visual-only until one node was saved.

## Validation
- npm run typecheck
- npm run lint
- npm run build
- npm test
- local header check confirmed /_next/static assets now return no-cache, must-revalidate in dev
- restarted next dev, cleared .next dev cache, hard-refreshed the browser, then verified /settings, /pipelines, and /contacts render without the Settings/Calendar/Search lucide module-factory errors

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -67,9 +67,9 @@ const nextConfig: NextConfig = {
    *   did nothing because the cache is server-side.
    *
    * Strategy:
-   *   - /_next/static/* — immutable for a year. Filenames are
-   *     content-hashed, so a new build produces new filenames; the
-   *     old ones are safe to keep indefinitely in caches.
+   *   - /_next/static/* — leave to Next. Turbopack dev chunks can go
+   *     stale if we force immutable caching here; Next already emits
+   *     the correct production headers for hashed assets.
    *   - /api/*          — no-store. API responses are per-user and
    *     must never be shared across requests at the edge.
    *   - Everything else — public, brief s-maxage + generous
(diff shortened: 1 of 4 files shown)


ArnasDon:
One small ask before merging: updateNodePositions routes through the dirtying setState, so just opening any pre-existing zero-positioned flow will flip dirty = true. The Save button will show as "needs saving" on every fresh open, and the beforeunload guard will fire if the user navigates away without saving — even though they didn't actually edit anything.

dhruv465:
Fixed in the latest push.

`updateNodePositions` now routes initial Dagre auto-layout hydration through `setStateRaw`, so opening a legacy all-zero flow no longer marks the editor dirty. Actual user drags still go through `updateNodePosition`, so Save/beforeunload behavior stays enabled for real layout edits.

Validation:
- `npm test -- src/components/flows/flow-editor-state.test.ts src/lib/flows/layout.test.ts`
- `npm test`
- `npm run typecheck`
- `npm run lint`
- `npm run build`

ArnasDon (approved):

PR #164, opened 2026-05-30
add offer catalog

dhruv465:
## Summary
- add a protected /products catalog page with create/list UI for offers
- add an offers table migration with authenticated grants and RLS scoped to each user
- add generic offer matching helper and tests for category, interest, city, budget, and metadata-based credit-card fees
- add Catalog to sidebar and page title mapping without changing existing navigation labels

## Why
This ports the catalog work from the CardReach branch onto current main without merging the larger CardReach changes. It keeps the current WhatsApp CRM modules intact while adding a reusable catalog for products, services, packages, eligibility rules, and commissions.

## Validation
- npm test -- src/lib/products/offers.test.ts
- npm test
- npm run typecheck
- npm run lint (0 errors, existing 20 warnings)
- npm run build
- browser check at http://localhost:3001/products showed Catalog nav/header, add-offer form, and existing offers rendering

## Notes
- supabase migration list --local could not run because local Postgres at 127.0.0.1:54322 is not running
- build still shows upstream's existing custom Cache-Control warning for /_next/static; separate PR #163 addresses that

### DIFF
--- a/src/app/(dashboard)/products/page.tsx
+++ b/src/app/(dashboard)/products/page.tsx
@@ -1,0 +1,313 @@
+'use client';
+
+import { useCallback, useEffect, useState } from 'react';
+import { toast } from 'sonner';
+import { BriefcaseBusiness, Loader2, Plus } from 'lucide-react';
+import { createClient } from '@/lib/supabase/client';
+import type { Offer } from '@/types';
+import { Button } from '@/components/ui/button';
+import { Input } from '@/components/ui/input';
+
+interface ProductForm {
+  name: string;
+  category: string;
+  provider: string;
+  price_amount: string;
+  fee_amount: string;
+  benefits: string;
+  min_budget: string;
+  city: string;
+  requirements: string;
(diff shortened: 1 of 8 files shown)


ArnasDon:
I'm going to close this without merging: wacrm is scoped specifically as a WhatsApp CRM, and the catalog doesn't connect to contacts, conversations, flows, or automations — merging it would add a parallel module that lives alongside the CRM rather than as part of it. A few of the design choices also reflect the CardReach origin (INR-only currency, credit-card-specific logic in minimumBudgetFor, timestamp-style migration name vs the sequential convention here) — not bad code, just designed for a different product.

wacrm is MIT-licensed, so this is exactly the kind of change that's perfect to maintain on your fork — pull in upstream changes whenever you want, and the catalog stays where it makes sense. If you ever want to land something narrower upstream (e.g. attaching a "product mentioned" to a contact, or a template variable for a product link), happy to discuss in an issue first.

PR #193, opened 2026-06-01
fix: resolve database migrations and account tenancy constraint issues

2527DC:
- Database Migrations: Replaced legacy uuid_generate_v4() with modern, built-in gen_random_uuid() across all SQL migrations to resolve extension search-path errors during deployment.
- Pipeline Creation: Fixed an RLS violation when seeding/creating pipelines by pulling accountId from the useAuth hook and including it in the insert payload.
- Flow Creation: Resolved a 500 API error in POST /api/flows by updating the requireUser auth guard to fetch the user's account_id and injecting it into the flow creation queries.

## Test plan


ArnasDon:
Thanks for the fix — the bugs are real (both pipelines and flows have account_id NOT NULL since migration 017, so inserts without it fail). A few things to address before merge:

Required changes:

 
- Flows route: don't reinvent requireUser — use the existing requireRole() / getCurrentAccount() helpers in src/lib/auth/account.ts. They already return { userId, accountId, supabase, ... } with typed errors.
- Pipelines seed race: seedAttempted.current = true is set before the seed runs, so fresh signups (where accountId is null on first render) permanently skip seeding even after the profile loads. Either gate the effect on accountId being truthy, or move that line inside the if (seeded) branch.
- Split the UUID swap into its own PR — it's unrelated to the tenancy fix and per CONTRIBUTING.md we keep one logical change per PR. Also drop the now-unused CREATE EXTENSION IF NOT EXISTS "uuid-ossp" in 001 if you keep that change.
- Rebase — branch is ~5 PRs behind main.

Nice to have: a smoke test for the flows POST account_id stamping.

2527DC:
### 🛠️ Required Changes Addressed

* **Flows API Route Auth Refactoring (`src/app/api/flows/route.ts`):**
  * Replaced the custom `requireUser` implementation with the existing centralized `getCurrentAccount()` / `requireRole()` auth helpers from `src/lib/auth/account.ts`.
  * Integrated `toErrorResponse(err)` to map authentication/authorization errors with proper typed statuses (401/403) and responses.

* **Pipelines Seed Race Fix (`src/app/(dashboard)/pipelines/page.tsx`):**
  * Gated the initial load and seeding `useEffect` on `accountId` being truthy, and added it to the dependency array.
  * This ensures fresh signups (where `accountId` is initially null) do not permanently skip seeding on subsequent renders after their profile loads.

* **Reverted UUID Generation Changes:**
  * Restored database migrations back to their original `uuid_generate_v4()` schemas. The update to `gen_random_uuid()` has been stripped out and will be submitted as a separate PR.

* **Added Flows API Smoke Tests (`src/app/api/flows/route.test.ts`):**
  * Added unit/integration tests covering:
    * Correct stamping of `user_id` and `account_id` when creating a new flow.
    * Correct propagation and stamping of `account_id` when cloning flow templates.
    * Expected HTTP responses for validation (name missing, invalid JSON) and authorization errors.

---

### 🧪 Verification Results

* **Unit/Integration Tests:** All 376 tests run and pass successfully via vitest (`npm run test`).
* **TypeScript Compilation:** All types compiled successfully without errors via tsc (`npm run typecheck`).

ArnasDon:
Closing as superseded — the account_id stamping for pipelines and flows landed independently in #210 (merged 2026-06-05), and this branch now conflicts on those same lines.

One piece is still worth keeping: the seed-race guard. #210 added the accountId check inside seedDefaultPipeline but left seedAttempted.current = true before the call, so a fresh signup whose profile resolves after first render still permanently skips default-pipeline seeding. Your top-level if (!accountId) return; + accountId dep fixes that. If you want, open a small PR off latest main with just that (and the flows getCurrentAccount refactor / tests if you like).

Thanks for the fix and the thorough follow-up.

PR #246, opened 2026-06-15
feat(contacts): CSV import tag assignment and preview improvements

iiamabhishek:
## Summary

Extends contact CSV import to support an optional `tags` column. Tag names from the file are assigned to newly imported contacts via `contact_tags`, matching the behavior of the manual contact form.

Also improves the import modal preview: colored tag badges, dynamic columns (hide empty `company`), truncated cells, and a wider layout for large files.

## Motivation

The import flow already handled `phone`, `name`, `email`, and `company`, but ignored tags entirely. Users exporting contacts from spreadsheets (often with a status/tag column) had to assign tags one-by-one after import.

### CSV parsing (`src/lib/contacts/parse-contact-csv.ts`)
- New shared parser for contact CSV files
- Optional `tags` column — comma/semicolon-separated values; quoted cells supported for multi-tag values (e.g. `"VIP, Lead"`)
- Returns metadata: `hasTagsColumn`, `hasCompanyColumn`
- Unit tests in `parse-contact-csv.test.ts`

### Tag resolution (`src/lib/contacts/resolve-import-tags.ts`)
- Resolves tag names to IDs (case-insensitive match against account tags)
- **Admin+**: auto-creates missing tags during import
- **Agent/viewer**: applies existing tags only; unknown names are skipped with a toast
- Batch `contact_tags` upsert after contacts are inserted (idempotent, ignores duplicates)

### Import modal (`src/components/contacts/import-modal.tsx`)
- Preview shows tag badges with account colors (or default blue for new tags)
- Tag summary pill (e.g. `1 tag · 57 contacts`)
- Hides `company` column when no company data in preview
- UI polish: wider modal, truncated filename/cells, scrollable table

## CSV format

```csv
phone,name,email,company,tags
+15551234567,Alice,alice@example.com,Acme,"VIP, Lead"
+15559876543,Bob,,,Customer

### DIFF
--- a/src/components/contacts/import-modal.tsx
+++ b/src/components/contacts/import-modal.tsx
@@ -8,7 +8,7 @@ import {
   isUniqueViolation,
   normalizeKey,
 } from '@/lib/contacts/dedupe';
-import { parseContactCsv } from '@/lib/contacts/parse-contact-csv';
+import { parseContactCsv, type ParsedContactRow } from '@/lib/contacts/parse-contact-csv';
 import {
   assignImportedContactTags,
   resolveImportTagIds,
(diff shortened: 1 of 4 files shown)


ArnasDon (changes requested):
Solid PR — extracting parseCSV and tag resolution into shared, tested modules fits the codebase well. Verified against schema/RLS/roles: no feature regressions, contact_tags upsert shape is correct, role gating matches RLS, and account_id is stamped correctly. A few things before merge:

1. (Blocker) Tag-assignment failure masks a successful import. assignImportedContactTags runs after contacts are committed but throws into the single catch → user sees "Import failed" and the list never refreshes, even though contacts imported. Wrap it:

try {
  tagsAssigned = await assignImportedContactTags(supabase, tagAssignments, tagIdByKey);
} catch {
  toast.warning('Contacts imported, but some tag assignments failed.');
}
2. (Please fix) resolveImportTagIds runs even with no tags — a transient SELECT error then aborts a tag-less import before any contact is inserted. Gate it on allTagNames.length.

3. Document the ordering invariant. inserted[j] ↔ chunk[j] only holds because a single INSERT preserves RETURNING order. Add a comment so it isn't silently broken by splitting the insert later.

4. tagsAssigned overcounts — it counts requested pairs, not actual inserts (ignoreDuplicates drops some silently). Note this in the function's doc comment.

5. Re-add the dropped step comments in handleImport (the dedupe / phone_normalized migration-022 / 23505 notes) — they documented non-obvious behavior.

Nits: preview hides company via AND but tags via OR (intentional? worth a comment); new files use double quotes vs .prettierrc singleQuote; description says tag handling "matches the manual form" but the form only selects existing tags, doesn't create them.

ArnasDon (approved):
Improvements looks good.

PR #279, opened 2026-06-22
feat: Notifications

AlexTPaul:
# Notifications

Agents receive in-app notifications when a teammate assigns them a conversation.

## Features

- **Notifications page** (`/notifications`) — lists the last 100 notifications in reverse-chronological order. Each item shows who assigned the conversation, the contact's name, and a relative timestamp. Clicking an item marks it read and navigates directly to that conversation in the inbox.
- **Mark all as read** — clears all unread items in a single action with optimistic UI so there's no visible delay.
- **Sidebar badge** — an unread count badge on the Notifications nav entry gives agents an at-a-glance signal when something needs attention.
- **Real-time updates** — new assignments appear instantly and badge counts stay in sync across tabs and devices via Supabase Realtime, without a page refresh.

## How notifications are created

Notifications are inserted exclusively by the `notify_conversation_assigned` database trigger (`supabase/migrations/023_notifications.sql`), which fires on `INSERT` or `UPDATE` of `conversations.assigned_agent_id`.

Rules the trigger enforces:

- No notification is created when `assigned_agent_id` is `NULL`.
- Self-assignments are silently skipped — the assigning agent is not notified.
- If notification creation fails for any reason the assignment itself is never blocked (`EXCEPTION WHEN OTHERS` guard with a `WARNING` log).

## Security

Row-level security on the `notifications` table ensures:

| Operation | Policy |
|-----------|--------|
| `SELECT`  | `auth.uid() = user_id` — agents read only their own notifications |
| `UPDATE`  | `auth.uid() = user_id` — agents may only mark their own notifications read |
| `INSERT`  | No client policy — rows are created only by the `SECURITY DEFINER` trigger |
| `DELETE`  | No client policy |

## Data model

```ts
interface Notification {
  id: string;
  account_id: string;
  user_id: string;          // recipient (assigned agent)
  type: 'conversation_assigned';
  conversation_id?: string;
  contact_id?: string;
  actor_user_id?: string;   // null when assigned by an automation/system
  title: string;
  body?: string;
  read_at?: string;         // null = unread
  created_at: string;
}
```

## Key files

| File | Purpose |
|------|---------|
| [src/app/(dashboard)/notifications/page.tsx](../src/app/(dashboard)/notifications/page.tsx) | Notifications page UI |
| [src/hooks/use-unread-notifications.ts](../src/hooks/use-unread-notifications.ts) | Hook that tracks the live unread count for the sidebar badge |
| [src/types/index.ts](../src/types/index.ts) | `Notification` and `NotificationType` type definitions |
| [supabase/migrations/023_notifications.sql](../supabase/migrations/023_notifications.sql) | Table, indexes, RLS policies, and assignment trigger |


ArnasDon (commented):
Thanks for this — really thorough PR description, clean RLS design, and nice touches like the optimistic UI and the `EXCEPTION WHEN OTHERS` guard so a notification failure can never block an assignment. A few things need to be addressed before this can merge, grouped by severity.

---

### 🔴 Blocking

**1. Migration number collides with an existing file.**
`supabase/migrations/023_notifications.sql` — `023` is already taken by `023_chat_media.sql`, and migrations currently run through `026_api_keys.sql`. Two files with the same prefix will break ordering / re-running migrations. Please rename to `027_notifications.sql` (and update the "migration 023" references in `src/types/index.ts` and the PR body).

**2. First line of the migration is invalid SQL.**
The file opens with a single dash:
```sql
- ============================================================
-- NOTIFICATIONS
```
That first line is `-` (minus operator), not `--` (comment) — Postgres will throw a syntax error and the whole migration fails to apply. Add the missing dash. Please run the migration against a clean DB to confirm it applies end-to-end.

**3. `use-unread-notifications.ts` reads `payload.old.read_at`, which isn't populated.**
By default a table's replica identity is the primary key, so realtime `UPDATE`/`DELETE` payloads only carry the PK in `payload.old` — every other column is `undefined`. So `const wasUnread = !oldRow.read_at` is always `true`, and the count will drift (e.g. any update to an already-read row decrements again). Note `useTotalUnread` deliberately never reads non-PK fields from `payload.old` for exactly this reason. Two options:
- Add `ALTER TABLE notifications REPLICA IDENTITY FULL;` to the migration so old values are present, **or**
- Drop the `wasUnread` logic and derive purely from `newRow.read_at` (updates here only ever set `read_at`).

---

### 🟡 Should fix

**4. Hardcoded `slate-*` / `white` / `red-*` colors break light mode.**
`notifications/page.tsx` uses `text-white`, `text-slate-400`, `bg-slate-900`, `border-slate-800`, `text-red-400`, etc. The app now has light + dark mode and the other dashboard pages use semantic tokens — please swap to `text-foreground`, `text-muted-foreground`, `bg-card`, `border-border`, `text-destructive`, and `bg-muted` so the page themes correctly. (Sidebar badge `text-primary-foreground` is already correct.)

**5. `markRead` swallows failures.**
`markAllRead` reloads on error, but `markRead` does the optimistic update and then fires the request with no error handling — if it fails the row stays "read" in the UI but unread in the DB. Mirror the `markAllRead` pattern (toast + `load()` on error).

**6. Page query isn't scoped to the active account.**
`page.tsx` selects all notifications for the user with no account filter. For a user who belongs to multiple accounts, this mixes notifications across accounts, whereas the rest of the app (inbox etc.) is scoped to the active account. Is that intended? If not, filter by the current `account_id`.

---

### ⚪ Nits

**7. Missing trailing newline** on `page.tsx`, `sidebar.tsx`, and `use-unread-notifications.ts` (the sidebar edit also stripped the existing file's final newline). Should be caught by lint/prettier.

**8. UPDATE RLS policy is column-wide.** `notifications_update` lets a user rewrite any column of their own row (title, conversation_id, …), not just `read_at`. Low risk since it's their own data, but if you want to harden it you could restrict to `read_at` via a trigger/check. Optional.

---

### Security review

Looks solid overall: the table is RLS-enabled, SELECT/UPDATE are scoped to `auth.uid() = user_id`, there's no client INSERT/DELETE policy, the trigger is `SECURITY DEFINER` with `SET search_path = public`, and realtime `postgres_changes` honors the SELECT policy so users only receive their own rows. No injection or XSS (all text is server-generated and React-escaped). The self-assignment skip and NULL-actor handling are correct. The only security-adjacent note is #8 above.

Nice work for a first contribution — once 1–3 are fixed this is close.

ArnasDon (changes requested):
A few things need addressing before we can merge, grouped by severity so it's easy to triage.

🔴 Blocking
1. Migration number collides with an existing file
supabase/migrations/023_notifications.sql — 023 is already used by 023_chat_media.sql, and our migrations currently run through 026_api_keys.sql. Two files sharing a prefix breaks ordering and re-running migrations. Please rename to 027_notifications.sql, and update the "migration 023" references in src/types/index.ts and the PR description to match.

2. First line of the migration is invalid SQL
The file opens with a single dash:

- ============================================================
-- NOTIFICATIONS
That first line is - (the minus operator), not -- (a comment) — Postgres throws a syntax error and the whole migration fails to apply. Add the missing dash. Please also run the migration against a clean database to confirm it applies end-to-end.

3. use-unread-notifications.ts reads payload.old.read_at, which isn't populated
By default a table's replica identity is its primary key, so realtime UPDATE/DELETE payloads only carry the PK in payload.old — every other column comes through as undefined. That means const wasUnread = !oldRow.read_at is always true, and the badge count will drift (e.g. any update to an already-read row decrements it again). Note that useTotalUnread deliberately never reads non-PK fields from payload.old for exactly this reason. Two ways to fix:

Add ALTER TABLE notifications REPLICA IDENTITY FULL; to the migration so old values are present, or
Drop the wasUnread branch and derive purely from newRow.read_at (updates here only ever set read_at, so this is simpler and sufficient).
🟡 Should fix
4. Hardcoded slate-* / white / red-* colors break light mode
notifications/page.tsx uses text-white, text-slate-400, bg-slate-900, border-slate-800, text-red-400, etc. The app now supports light + dark mode and the other dashboard pages use semantic tokens. Please swap to text-foreground, text-muted-foreground, bg-card, border-border, text-destructive, and bg-muted so the page themes correctly. (The sidebar badge's text-primary-foreground is already right.)

5. markRead swallows failures
markAllRead reloads on error, but markRead does the optimistic update and fires the request with no error handling — if it fails, the row stays "read" in the UI but unread in the DB. Mirror the markAllRead pattern: toast + load() on error.

6. Page query isn't scoped to the active account
page.tsx selects all notifications for the user with no account filter. For a user who belongs to multiple accounts, this mixes notifications across accounts, whereas the rest of the app (inbox, etc.) scopes to the active account. Is that intentional? If not, filter by the current account_id.

⚪ Nits
7. Missing trailing newline on page.tsx, sidebar.tsx, and use-unread-notifications.ts (the sidebar edit also stripped that file's existing final newline). Lint/prettier should catch these.

8. UPDATE RLS policy is column-wide — notifications_update lets a user rewrite any column of their own row (title, conversation_id, …), not just read_at. Low risk since it's their own data, but if you want to harden it you could restrict updates to read_at. Optional.

AlexTPaul:
# Notifications Feature — PR Fix Summary

## Feature Overview

Added a notifications system that alerts agents when a conversation is assigned to them. Includes a database table, realtime updates, a sidebar badge, and a dedicated `/notifications` page.

---

## 🔴 Blocking Fixes

### Migration renamed: `023_notifications.sql` → `027_notifications.sql`

The original file shared the `023` prefix with `023_chat_media.sql`, breaking migration ordering. Renamed to follow sequentially after `026_api_keys.sql`. Updated the reference comment in `src/types/index.ts`.

### First-line SQL syntax error fixed

The migration opened with a single `-` (the minus operator) instead of `--` (a comment), causing Postgres to throw a syntax error and reject the entire migration.

### Realtime payload fix — `REPLICA IDENTITY FULL` added

By default Postgres only includes the primary key in `payload.old` for realtime UPDATE events. This caused `use-unread-notifications.ts` to read `oldRow.read_at` as `undefined`, making the unread badge count drift on every update. Fixed by:

- Adding `ALTER TABLE notifications REPLICA IDENTITY FULL;` to the migration
- Simplifying the UPDATE handler to derive purely from `newRow.read_at` (since updates only ever set `read_at`, this is sufficient and more robust)

---

## 🟡 Should-Fix Issues

### Semantic color tokens — light mode support

`notifications/page.tsx` used hardcoded `text-white`, `text-slate-*`, `bg-slate-900`, `text-red-400`, etc. Replaced with semantic tokens so the page themes correctly in both light and dark mode.

| Before | After |
|--------|-------|
| `text-white` | `text-foreground` |
| `text-slate-400` | `text-muted-foreground` |
| `bg-slate-900` | `bg-card` |
| `border-slate-800` | `border-border` |
| `bg-slate-800` | `bg-muted` |
| `text-red-400` | `text-destructive` |

### `markRead` error handling

Previously did the optimistic update and fired the request with no error handling — a failed request left the row visually read but unread in the DB. Now mirrors `markAllRead`: shows a toast and calls `load()` to revert the optimistic state on failure.

### Page scoped to active account

The notifications query had no `account_id` filter, mixing notifications across all accounts for multi-account users. Added `.eq("account_id", accountId)` using the `accountId` from `useAuth()`, matching how inbox and other pages scope their data.

---

## ⚪ Nits

### Trailing newlines

Added missing trailing newlines to `sidebar.tsx` and `use-unread-notifications.ts`.

### UPDATE RLS restricted to `read_at`

The original `notifications_update` policy allowed clients to rewrite any column of their own rows. Hardened using Postgres column-level privileges:

```sql
REVOKE UPDATE ON notifications FROM authenticated;
GRANT UPDATE (read_at) ON notifications TO authenticated;
```

This is the correct approach — the `WITH CHECK` clause alone cannot restrict which columns are written.

---

## Files Changed

| File | Change |
|------|--------|
| `supabase/migrations/027_notifications.sql` | New — renamed from 023, fixed SQL, added `REPLICA IDENTITY FULL`, column-level UPDATE grant |
| `supabase/migrations/023_notifications.sql` | Deleted |
| `src/types/index.ts` | Updated migration reference comment to 027 |
| `src/hooks/use-unread-notifications.ts` | Simplified UPDATE handler; added trailing newline |
| `src/app/(dashboard)/notifications/page.tsx` | Semantic colors, `markRead` error handling, account scoping |
| `src/components/layout/sidebar.tsx` | Added trailing newline |

ArnasDon (approved):

PR #308, opened 2026-06-26
fix(settings): preserve WhatsApp credential drafts

### LINKED ISSUE #307: [bug] WhatsApp settings form clears unsaved fields after tab focus
### What happened?

Unsaved values in the WhatsApp settings form can be cleared after switching to another browser tab and coming back.

This is easy to hit while copying Meta credentials from another tab. The user types a value into one of the WhatsApp credential inputs, changes tabs to copy the next value, then returns and the input has been reset to the stored database value or an empty string.

### Steps to reproduce

1. Sign in locally.
2. Go to `/settings?tab=whatsapp`.
3. Type a value into one of the API credential fields, for example `Phone Number ID`.
4. Switch to another browser tab.
5. Return to the wacrm tab.
6. Check the value that was typed but not saved yet.

### What did you expect?

Unsaved form input should stay in place while the settings panel remains mounted. Switching browser tabs should not discard draft credential values.

### Commit / version

`b2ed58a`

### Where is it running?

Local dev (npm run dev)

### Logs / screenshots

No console error is required to reproduce it. The field is reset after the tab is focused again.

bellopushon:
## Summary

Keep unsaved WhatsApp credential fields from being overwritten when the browser tab regains focus.

## What changed

- Track the account id that has already loaded the WhatsApp config.
- Skip repeated same-account config hydration so auth/profile refreshes do not reset in-progress form input.
- Still allow explicit reloads after save or registration verification.

## Test plan

- [x] `npm run lint` — no new errors beyond the pre-existing warning backlog.
- [x] `TZ=UTC npm test` passes.
- [ ] Browser repro not re-run here; this change is limited to the rehydration path that was resetting the controlled inputs.

## Related

Closes #307

### DIFF
--- a/src/components/settings/whatsapp-config.tsx
+++ b/src/components/settings/whatsapp-config.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useEffect, useState, useCallback } from 'react';
+import { useEffect, useRef, useState, useCallback } from 'react';
 import { toast } from 'sonner';
 import {
   Eye,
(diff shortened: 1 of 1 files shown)


Darshan072 (approved):

ArnasDon:
Fix merged with other PR. Closing this.

Issue #173, opened 2026-05-31
[feat] Make UI text translatable so forks can change language without conflicts


Am1n3:
### What's the problem?

Every user-facing string is hardcoded in English, inline across ~40–50 component files (sidebar, auth pages, settings, inbox, etc.). There is no separate place for translations.

That means any forker deploying for a non-English audience has to edit those component files directly — and those are exactly the files you keep improving. So every time they pull your updates, the changes will cause merge conflicts.

Right now you cannot do both: translate the app AND keep getting updates. You have to choose one.

### What would you like to see?

A simple way to translate the text. Instead of writing English directly in the code, the app would read each text from a language file (for example messages/en.json). English stays the default.

For someone who forks the template:
- By default nothing changes. English is still the default. The app looks and works exactly the same. This is only an addition, and it is optional.
- To add a language, they just add one new file (for example messages/fr.json) and pick a default language. A new file does not cause conflicts. So they get a translated app AND clean updates from you.

The main benefit is for the next person who forks: they can translate by adding one file, instead of rewriting code and fighting conflicts later.

I am happy to do this work myself. I would send it as small, easy-to-review pull requests: first the setup plus one example screen, then the rest in small groups. Each one keeps the same behavior and passes CI. Unless you prefer it as one PR with all the changes ?

### Alternatives you considered

- Edit the text directly in the fork: it works, but it causes merge conflicts on every update. That is the problem this idea removes.
- A layer that replaces English text on screen at runtime: it needs no change from you, but it is not reliable. It goes back to English when you change the wording, and it does not handle text with numbers or plurals well.
- Keep it only in the fork: it works, but every person who forks has to solve the same problem again. Adding it here solves it once, for everyone.

### Scope

None

ivanmercedes:
I've started working on this. I just opened an initial PR (#328 ) containing the base next-intl configuration and the migration for the Login page. I will continue working on translating the rest of the application incrementally.

Am1n3:
Thanks for addressing this issue.

PR #328, opened 2026-07-02
feat: integrate next-intl for i18n support

ivanmercedes:
## Summary

This PR introduces the foundational support for internationalization (i18n) using next-intl and migrates the initial login page to use the new translation system instead of hardcoded strings. This is the first step towards full multi-language support.

## What changed

- Added next-intl dependency to the project.
- Configured next.config.ts with the next-intl plugin.
- Created the global configuration file src/i18n/request.ts.
- Injected NextIntlClientProvider into src/app/layout.tsx.
- Created the initial English dictionary at messages/en.json.
- Refactored src/app/(auth)/login/page.tsx to use dynamic translation keys.
- Added NEXT_PUBLIC_APP_LOCALE to .env.local.example.

## Test plan

## Related

Part of #173

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -42,6 +42,9 @@ META_APP_SECRET=your-meta-app-secret
 # or a background worker that has no incoming request.
 NEXT_PUBLIC_SITE_URL=https://crm.example.com
 
+# Default language locale (e.g. en, es)
+NEXT_PUBLIC_APP_LOCALE=en
+
 # ============================================================
 # OPTIONAL — only needed if you use the feature.
 # ============================================================
--- a/.env.local.example
+++ b/.env.local.example
@@ -42,7 +42,7 @@ META_APP_SECRET=your-meta-app-secret
 # or a background worker that has no incoming request.
 NEXT_PUBLIC_SITE_URL=https://crm.example.com
 
-# Default language locale (e.g. en, es)
+# Default language locale (e.g. en)
 NEXT_PUBLIC_APP_LOCALE=en
 
 # ============================================================
--- a/messages/en.json
+++ b/messages/en.json
@@ -1,0 +1,17 @@
+{
+  "LoginPage": {
+    "titleAccept": "Sign in to accept",
+    "titleWelcome": "Welcome back",
+    "descAccept": "Sign in and we'll take you to the invitation.",
+    "descWelcome": "Sign in to your account",
+    "emailLabel": "Email",
+    "emailPlaceholder": "you@example.com",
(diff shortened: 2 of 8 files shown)


ArnasDon (changes requested):
Reviewed and validated locally. Typecheck passes, and the build succeeds with the default en locale, so the core change is solid and the login page migration is clean — all 13 translation keys match en.json.

One blocker before this can merge:

src/i18n/request.ts crashes for any locale other than en. The dynamic import(\../../messages/${locale}.json`)has no fallback, anden.jsonis the only dictionary that exists. This isn't hypothetical —.env.local.exampleliterally suggestsesas a value ("e.g. en, es"), so anyone following the example and settingNEXT_PUBLIC_APP_LOCALE=esgets a hard failure:Module not found: Can't resolve '../../messages/es.json'`. Because Turbopack resolves it at build time, it breaks the whole build, not just one request.

Please either:

wrap the import and fall back to en when the file is missing (or gate on a known-locales list), or
drop the es hint from .env.local.example until a second dictionary actually exists.
Minor (optional): NextIntlClientProvider gets messages but no locale prop — works today, but passing locale explicitly is more correct for client-side date/number formatting down the line.

ivanmercedes:
> Reviewed and validated locally. Typecheck passes, and the build succeeds with the default en locale, so the core change is solid and the login page migration is clean — all 13 translation keys match en.json.
> 
> One blocker before this can merge:
> 
> src/i18n/request.ts crashes for any locale other than en. The dynamic import(../../messages/${locale}.json. Because Turbopack resolves it at build time, it breaks the whole build, not just one request.`)has no fallback, anden.jsonis the only dictionary that exists. This isn't hypothetical —.env.local.exampleliterally suggestsesas a value ("e.g. en, es"), so anyone following the example and settingNEXT_PUBLIC_APP_LOCALE=esgets a hard failure:Module not found: Can't resolve '../../messages/es.json'`
> 
> Please either:
> 
> wrap the import and fall back to en when the file is missing (or gate on a known-locales list), or drop the es hint from .env.local.example until a second dictionary actually exists. Minor (optional): NextIntlClientProvider gets messages but no locale prop — works today, but passing locale explicitly is more correct for client-side date/number formatting down the line.

Done! Added the fallback logic for missing dictionaries, removed the es hint from the env example, and added the locale prop.

ArnasDon (approved):

ivanmercedes:
Synced package-lock.json to fix npm ci build failures. @ArnasDon

PR #332, opened 2026-07-02
fix(uazapi): record app-sent messages and fix media download field name

jonesfernandess:
## Summary
- The Uazapi webhook discarded every `fromMe` event, assuming it was always an echo of a CRM-sent message. Messages sent from the linked phone's WhatsApp app (outside the CRM) never got recorded, so conversation history was incomplete.
- `fromMe` events now resolve/create the contact and conversation, dedupe against the CRM's own send-path insert by `message_id`, and insert as `sender_type: 'agent'` — without triggering unread bumps, flows, automations, or funnel entry (kept customer-reply-only).
- Fixes phone resolution for `fromMe` events: `sender`/`sender_pn` identify the account owner in that case, not the conversation partner — using them there was filing every app-sent message under a "chat with myself" contact instead of the real recipient. The partner's number now comes from `chatid`/`chat.phone` when `fromMe` is true.
- Unrelated fix bundled in: `/message/download` reads Uazapi's actual response field `base64Data` (not `base64`), which was silently failing every media download.

## Test plan
- [x] `tsc --noEmit` passes
- [x] Manually verified via ngrok tunnel: sent a message from the WhatsApp app to a contact — message now appears in the correct conversation in the CRM, not "chat with myself"
- [ ] Verify a message sent through the CRM itself still shows once (no duplicate from the webhook echo)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,13 @@
+# O que NÃO entra no contexto de build.
+# Atenção: .env.local NÃO está aqui — ele é necessário no build (NEXT_PUBLIC_*).
+node_modules
+.next
+.git
+.github
+npm-debug.log*
+*.log
+.DS_Store
+Dockerfile
+.dockerignore
+docker-stack.yml
+deploy.sh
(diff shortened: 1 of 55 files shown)


jonesfernandess:
Opened against the wrong repo by mistake (should target jonesfernandess/wacrm, my fork). Closing — recreating correctly.

PR #362, opened 2026-07-09
feat: translate frontend to Spanish

Zarko-cl:
Adds messages/es.json (full next-intl translation) and switches the default locale to es. Also translates hardcoded UI text that bypassed next-intl across app pages and components, plus user-facing strings in lib/ that render in the UI: dashboard activity feed, currency and template-status labels, flow/automation quick-start templates, flow/automation validation messages, interactive-payload validation, and theme names.

Updates the handful of unit tests that asserted exact English message strings from the translated validation/interactive modules.

## Test plan


ivanmercedes:
Nice contribution 👍

One thing I noticed is that this PR also translates hardcoded strings inside components and changes the default locale, not only `es.json`.

Personally, I think it would be cleaner to keep translation PRs limited to language files and i18n keys, and avoid modifying global component text or the default locale.

ArnasDon (commented):
Thanks for this — the `es.json` dictionary is solid work and I'd like to get language support in. But there are two blockers, both the same principle: **a new language should be additive and opt-in, never the new default.**

1. **Don't switch the default locale.** Flipping `NEXT_PUBLIC_APP_LOCALE` to `es` makes every fresh install and forker boot in Spanish. Keep the default `en` and document `es` as an available locale.
2. **Most strings are hardcoded Spanish instead of going through next-intl.** With `en` still the default and the fallback, English is now broken — half the UI is driven by `es.json`, half by hardcoded Spanish (note `en.json` wasn't touched). And the translated `lib/` validators are returned by API routes, so this leaks Spanish into API responses for all consumers (incl. the MCP server).

The mergeable version: keep `es.json`, leave the default on `en`, move any UI text that bypasses i18n into `t()` keys in **both** dictionaries, and revert the `lib/` + test changes (English stays canonical there). See inline notes.

.env.local.example
```diff
@@ -42,8 +42,8 @@ META_APP_SECRET=your-meta-app-secret
 # or a background worker that has no incoming request.
 NEXT_PUBLIC_SITE_URL=https://crm.example.com
 
-# Default language locale (e.g. en)
-NEXT_PUBLIC_APP_LOCALE=en
+# Default language locale (e.g. es)
+NEXT_PUBLIC_APP_LOCALE=es
```
ArnasDon, line 46: Please keep this on `en`. Changing the committed default makes every new install and every forker who hasn't set the var boot in Spanish — the language should be opt-in. Document `es` as an available locale instead.

src/app/(auth)/signup/page.tsx
```diff
@@ -49,12 +49,12 @@ function SignupPageInner() {
     setError(null);
 
     if (password !== confirmPassword) {
-      setError("Passwords do not match");
+      setError("Las contraseñas no coinciden");
```
ArnasDon, line 52: This hardcodes Spanish and bypasses next-intl, so it shows even under the default/fallback `en`. Please route these through `t()` with keys in both `en.json` and `es.json`. Same pattern applies across the other component files in this PR.

src/lib/whatsapp/interactive.ts
```diff
@@ -113,16 +113,16 @@ export function validateInteractivePayload(
   payload: unknown,
 ): InteractiveValidation {
   if (!payload || typeof payload !== 'object') {
-    return fail('Interactive message payload is required.')
+    return fail('Se requiere el contenido del mensaje interactivo.')
```
ArnasDon, line 116: These validator strings are returned in API responses (`/api/v1/messages`, `/api/quick-replies`, `/api/automations`, automation engine) via `NextResponse.json({ error: result.error })`, not just shown in the UI. Hardcoding Spanish here changes the API contract for every consumer (incl. the MCP server) with no locale negotiation. Please keep `lib/` messages English as the canonical output and translate at the UI layer.

ArnasDon:
Closing. There's real work here — ~70 files of hardcoded strings pulled into next-intl — but it can't land in this shape.

### The blocker

```diff
-# Default language locale (e.g. en)
-NEXT_PUBLIC_APP_LOCALE=en
+# Default language locale (e.g. es)
+NEXT_PUBLIC_APP_LOCALE=es
```

Switching the template's default language to Spanish changes what every forker gets out of the box. English stays the default and the fallback upstream; a fork picking `es` is exactly what `NEXT_PUBLIC_APP_LOCALE` exists for — one line in your own `.env.local`, no PR needed.

### Also in the way

- **Conflicts with `main`** in three files, and it's a month stale (9 July).
- **`messages/es.json` (+1799)** is a new locale catalogue nobody upstream can review or keep accurate. It also puts a standing tax on every future string: en, ko, *and* es or CI breaks.
- **Translated `lib/` strings.** Moving validation messages, interactive-payload errors, currency and template-status labels into the catalogue means the unit tests asserting on them had to be rewritten — you noted this in the body. Some of those strings are closer to protocol errors than UI copy, and translating them makes the failure modes harder to search for.

### What would land

The genuinely useful half is the same as in #376: hardcoded UI text that bypasses next-intl is a real defect regardless of which languages ship. A PR that pulls those strings into `messages/en.json` — with matching `ko.json` keys, English still the default, no `es.json` — is a clear improvement and I'd review it.

Worth knowing that #376 covers overlapping ground with the same conclusion, so it's worth coordinating if you both resend.

One adjacent thing you may hit: I fixed 12 strings in #477 that were rendering as their own keypath, because `{{1}}` and raw HTML aren't valid ICU and were being read with plain `t()` rather than `t.raw()`. If you re-extract strings, that test (`src/i18n/icu-safety.test.ts`) will catch the same mistake early.

PR #376, opened 2026-07-12
feat: multi-language support (pt/es) + internationalize remaining strings

### LINKED ISSUE #173: [feat] Make UI text translatable so forks can change language without conflicts

### LINKED ISSUE #375: Multi-language support (pt/es) + internationalize remaining hardcoded strings
### Proposal

Ship first-class multi-language support. Today `src/i18n/request.ts` already selects a locale from `NEXT_PUBLIC_APP_LOCALE` and falls back to English, but (a) only `messages/en.json` exists and (b) a number of screens bypass next-intl with hardcoded English, so they can't be localized at all.

This proposes:

1. Add `messages/pt.json` (Brazilian Portuguese) and `messages/es.json` (Latin American Spanish), key-for-key identical to `en.json`.
2. Internationalize ~250 hardcoded strings across ~30 files (Notifications, AI agent, Signup, Forgot password, Join, Flows header, interactive message builder, flow/automation template galleries, theme picker, scattered toasts/placeholders/aria-labels) so the whole UI follows the selected locale.
3. Fix a few pre-existing i18n bugs that render raw keys **even in English**:
   - missing `Settings.sections.quick-replies` label,
   - Overview role chip using `useTranslations('roles')` instead of `'Settings.roles'`,
   - the API-key `scopesHint` message embedding `<code className>` (next-intl rich tags must be attribute-less),
   - Base UI `<SelectValue />` triggers showing the raw enum value instead of the translated option label.

It stays additive: locale is still one global value, English remains the default and fallback, and data-definition files keep their English seed copy.

I understand the template is intentionally English-only and single-locale, and that translations may be considered fork territory — opening this first per CONTRIBUTING to check alignment. The bug fixes in point 3 are worth landing regardless of the translation decision; happy to split them into a separate English-only PR if you'd prefer to take just those.

opastorello:
## Summary

Ship full multi-language support: add pt/es catalogs, internationalize the remaining hardcoded strings, and fix several i18n leaks (some visible even in English). The locale stays a single global value driven by `NEXT_PUBLIC_APP_LOCALE`; English remains the default and fallback.

## What changed

- **New catalogs:** `messages/pt.json` (pt-BR) and `messages/es.json` (es-419), key-for-key identical to `en.json` (1683 leaf keys each; ICU/interpolation arguments preserved).
- **~250 hardcoded strings internationalized** across ~30 files: Notifications, AI agent (page + Playground + Usage), Signup, Forgot password, Join/invite, Flows header, interactive message builder/preview, flow & automation template galleries, theme picker, and scattered toasts/placeholders/aria-labels.
- **Bug fixes (raw keys shown even in English):** missing `Settings.sections.quick-replies`; Overview role chip scope `'roles'` -> `'Settings.roles'`; `scopesHint` `<code className>` -> attribute-less `<code>`.
- **Base UI Select triggers:** 14 bare `<SelectValue />` now render the selected option's translated label instead of the raw enum value.
- **API scope descriptions** moved to `Settings.apiKeys.scopeDesc.*`.

Data files (flow/automation templates, themes) keep English name/description as seed fallback; galleries resolve display copy via `useTranslations` keyed by each item's stable id.

## Test plan

- [x] `npm run lint` — 0 errors.
- [x] en/pt/es catalogs verified key-for-key identical; ICU argument parity checked per-string (0 mismatches); all `useTranslations` scopes resolve to a catalog namespace (guards against raw-key leaks).
- [x] Deployed with `NEXT_PUBLIC_APP_LOCALE=pt` and spot-checked the converted screens.

## Related

Closes #375. Also resolves #173 (make UI text translatable without fork conflicts).


ivanmercedes:
Great work on this! 

One small note: this PR also appears to resolve issue #173, so it might be worth adding Closes #173

opastorello:
Thanks @ivanmercedes! Good catch — updated the description to reference #173. This PR follows exactly that shape: English stays the default and fallback, and adding a locale is just one new `messages/<locale>.json` file, so forks can translate without fighting merge conflicts.

ArnasDon:
Reviewed the branch locally — key parity (1683 keys × en/es/pt), ICU argument parity, `typecheck`, and `lint` (0 errors) all check out, and `request.ts` loads the new catalogs correctly. A couple of small things worth improving:

**1. Missing `t` in some `useCallback` deps.** A few converted callbacks now call `t()` but don't list it in their dependency array, unlike the other files in this PR — it adds new `react-hooks/exhaustive-deps` warnings:
- `members-tab.tsx` (`load`, ~L177)
- `whatsapp-config.tsx` (~L166)
- `quick-replies-manager.tsx` (`save` L113, `remove` L125)

Harmless with a fixed global locale, but worth aligning for consistency (or intentionally leaving them all out).

**2. Two pre-existing raw-key leaks not covered** (optional, since this PR is cleaning up i18n leaks anyway):
- `automation-builder.tsx:1157` — `t("delete", { defaultValue: "Delete" })`
- `contact-detail-view.tsx:471` — `t("tabs.tags", { fallback: "Tags" })`

next-intl ignores `defaultValue`/`fallback`, and neither key exists, so both render the raw key path. Adding `Automations.builder.delete` and `Contacts.detailView.tabs.tags` (× 3 locales) would fix them.

Neither blocks merge — otherwise LGTM.

opastorello:
Thanks for the thorough review @ArnasDon! Both points addressed in d539518:

1. **`useCallback` deps** — added `t` to the dependency arrays in `members-tab.tsx`, `whatsapp-config.tsx`, and `quick-replies-manager.tsx` (both `save` and `remove`), matching the convention used elsewhere in the PR. The three `exhaustive-deps` warnings are gone.
2. **Raw-key leaks** — added `Automations.builder.delete` and `Contacts.detailView.tabs.tags` to all three catalogs (en/pt/es, now 1685 keys each, still in full parity) and dropped the ignored `defaultValue`/`fallback` options from the call sites.

`typecheck` and `lint` (0 errors) still clean.

ArnasDon:
This is the most carefully done of the i18n PRs in the queue — key-for-key parity across catalogues, English kept as default and fallback, and real bugs found along the way. I'm still closing it, but I want to be specific about why, because a rebased version of part of it would land.

### It can't merge as-is

**1. It breaks the locale parity test.** The branch predates `messages/ko.json` — it isn't on the branch at all — while `en.json` grows from 1447 to 1685 keys. After a merge, `ko.json` (still 1447 keys, inherited from `main`) would be missing **257** of them:

```
Sidebar.primaryNav
Inbox.messageThread.toastSendFailed
Inbox.messageThread.toastSendTemplateFailed
…
```

`src/i18n/messages.test.ts` fails on drift in either direction, so `npm test` goes red. Every new `en.json` key needs a `ko.json` counterpart in the same PR.

**2. Three-file conflict with `main`**, and it's a month stale — `main` has moved a fair bit since 12 July.

**3. Scope.** Two brand-new locale catalogues at ~1683 keys each is a feature, and one nobody upstream can review for correctness or maintain. Every future string then needs pt and es alongside en and ko, or CI breaks.

### The part I'd take today

Buried in here is a set of genuine correctness fixes that are independent of pt/es:

- **~250 hardcoded strings internationalized** across Notifications, AI agent, Signup, Forgot password, Join/invite, Flows header, the interactive message builder, template galleries and assorted toasts. That's straightforwardly fixing something the template does wrong, and it's welcome.
- **Missing key `Settings.sections.quick-replies`** — renders as a raw keypath in English today.
- **14 bare `<SelectValue />` triggers** showing raw enum values instead of translated labels.

Two of your other finds have already landed independently: the `scopesHint` `<code className>` attribute and the `'roles'` → `'Settings.roles'` scope fix.

And FYI, since it's adjacent to what you were working in — I merged #477 today fixing 12 strings that render as their own keypath because `{{1}}` and raw HTML can't be parsed as ICU and were being read with plain `t()` instead of `t.raw()`. Some of those are in files you touch here.

### If you want to resend

One PR against current `main`: the hardcoded-string extraction, the `quick-replies` key, and the `SelectValue` labels — with **en + ko both updated** for every new key, and no pt/es. That's reviewable and it's a clear improvement. The pt/es catalogues belong in a fork, where they can be kept accurate by people who actually speak the languages.

Thanks for the care that went into this — the parity discipline and the note about data files keeping English seeds as fallback both showed you'd thought it through.

PR #397, opened 2026-07-17
Add Spanish translations and a per-user language switcher

yzz0f:
## Summary
- Add `messages/es.json` — full Spanish (Latin American) translation, kept 1:1 with every key in `messages/en.json`.
- Wire ~10 components that were rendering hardcoded English text directly instead of using `next-intl` (signup, forgot-password, join/[token] invite page, Agents, Notifications, dashboard shell, Flows editor header, AI Playground, AI usage card, quick replies manager, interactive message preview).
- Add a per-user language switcher in Settings → Appearance (`src/components/settings/locale-switcher.tsx`).
- `src/i18n/request.ts` now resolves the locale per request: user cookie (set by the switcher) → browser `Accept-Language` → `NEXT_PUBLIC_APP_LOCALE` env var → `en`. Previously the whole deployment was pinned to a single locale via that env var with no per-user override.

## Test plan
- [x] `messages/en.json` and `messages/es.json` have identical key sets (verified with a script — 0 missing on either side)
- [x] `npx tsc --noEmit` clean on all touched `src/` files
- [x] `eslint` clean on all touched files
- [x] `next build` compiles successfully (the only remaining build error is in `mcp-server/`, a separate unrelated package with missing local deps, pre-existing on `main`)
- [ ] Manual check in a running dev environment: toggle the language switcher in Settings → Appearance and confirm the UI updates without a hard reload


yzz0f:
Wrong target — recreating against my fork instead.

Issue #243, opened 2026-06-12
[feat] Add theme switcher (light/dark mode toggle)


itcastleltd:
### What's the problem?

I'm trying to use the CRM during the day, but the current dark mode colors are too harsh and cause eye strain and headaches. The template makes it harder because there is no way to switch to a light theme, forcing me to use a color scheme that is physically uncomfortable for extended use. 

### What would you like to see?

As a user, I want a simple toggle button to switch the application between light mode and dark mode. This allows me to choose the theme that is most comfortable for my eyes, so I can work without getting headaches from the dark mode.

### Scope

Small (a couple of hours)

PR #402, opened 2026-07-17
feat(theme): add support for light theme and mode selector

DibashSarkar:
## Description

This PR introduces support for Light Mode across the application, adding a theme selector to allow users to switch between Light, Dark, and System preferences.

*Note: This is part 1 of splitting up the previous monolithic PR to allow for easier, safer reviews.*

### Changes
- Configured Tailwind CSS to support `dark:` variants properly alongside light mode defaults.
- Updated `globals.css` with a complete set of light and dark CSS variables for the color palette.
- Added a `ThemeToggle` component to the user settings/header for seamless switching.
- Ensured all existing components (buttons, cards, text, and inputs) have appropriate color contrast and styling in both themes.
- Replaced hardcoded color values with semantic Tailwind classes (e.g., `bg-background`, `text-foreground`).

### Why this is needed
To improve accessibility and provide a better user experience for those who prefer working in light mode environments.

### Testing
- [x] Verified manual switching between Dark and Light mode.
- [x] Verified System mode respects OS-level theme preferences.
- [x] Tested across major views (dashboard, settings, webhooks) to ensure no visual regressions or illegible text.

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -1,6 +1,7 @@
 @import "tailwindcss";
 @import "tw-animate-css";
 @import "shadcn/tailwind.css";
+@import "./light.css";
 
 @custom-variant dark (&:is(.dark *));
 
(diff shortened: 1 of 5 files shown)


ArnasDon:
Closing — light mode already exists upstream. `MODES = ["light", "dark"]` is in `src/lib/themes.ts` and the `html[data-mode="light"]` block is at `src/app/globals.css:119` (added in #243). The new `light.css` re-declares that same selector but is imported above it, so source order means the existing block wins.

Two other things if you revisit this: registering `light` in `THEME_IDS` mixes the accent and light/dark dimensions that `themes.ts` keeps separate, and there's no `html[data-theme="light"]` block to back it. The `layout.tsx` description also became "Self-hostable Journey Tank WP CRM", which looks like it came along from your fork.

PR #404, opened 2026-07-17
fix(whatsapp): resolve issues in template-sending paths and media headers

DibashSarkar:
## Description

This PR addresses several critical bugs in the WhatsApp message sending paths, specifically around how templates with media headers are processed and dispatched to the Meta API.

*Note: This is part 3 of splitting up the previous monolithic PR to allow for easier, safer reviews.*

### Changes
- Fixed serialization and attachment issues when sending templates containing rich media headers (images/documents).
- Improved error handling and validation in the outgoing message payload construction paths.
- Ensured that `wamid` mapping and message status callbacks correctly associate with the dispatched template.

### Why this is needed
To ensure reliability in outbound messaging. Prior to these fixes, certain templates with media components would fail to send or be malformed, resulting in API rejections from Meta.

### Testing
- [x] Verified sending standard text templates.
- [x] Verified sending templates with media headers (images/documents) successfully deliver to the WhatsApp client.
- [x] Automated tests (`npm run test`) pass without regressions.

### DIFF
--- a/src/app/api/whatsapp/webhook/route.ts
+++ b/src/app/api/whatsapp/webhook/route.ts
@@ -57,6 +57,7 @@ interface WhatsAppMessage {
     button_reply?: { id: string; title: string }
     list_reply?: { id: string; title: string; description?: string }
   }
+  button?: { payload?: string; text?: string }
   /** Present when the customer swipe-replies to one of our messages. */
   context?: { id: string }
 }
@@ -653,7 +654,9 @@ async function processMessage(
     ? message.type
     : message.type === 'sticker'
       ? 'image'   // stickers are images
-      : 'text'    // reaction, unknown → text fallback
+      : message.type === 'button'
+        ? 'interactive' // quick replies are interactive button replies
+        : 'text'    // reaction, unknown → text fallback
 
   // Determine whether this is the contact's very first inbound message
   // BEFORE we insert, so the count is accurate. Covers the case where
@@ -961,6 +964,17 @@ async function parseMessageContent(
         }
       }
       return { ...empty, contentText: '[Interactive reply]' }
+    }
+
+    case 'button': {
+      if (message.button) {
+        return {
+          ...empty,
+          contentText: message.button.text || null,
+          interactiveReplyId: message.button.payload || null,
+        }
+      }
+      return empty
     }
 
     default:
--- a/src/components/contacts/contact-detail-view.tsx
+++ b/src/components/contacts/contact-detail-view.tsx
@@ -330,6 +330,11 @@ export function ContactDetailView({
   ) {
     if (!contactId) return;
     setSendingTemplate(true);
+    const renderedBody = template.body_text.replace(/\{\{(\d+)\}\}/g, (_, raw) => {
+      const idx = Number(raw) - 1;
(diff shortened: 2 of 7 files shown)


ArnasDon:
Closing this one, but not because it's all wrong — there's a real fix buried in here that I'd like back as its own PR. Details below.

### The good part

The `message.type === 'button'` handling in `webhook/route.ts` is a genuine gap in upstream. Meta sends template quick-reply taps as `type: 'button'` with `{ payload, text }`, which is a different shape from the `type: 'interactive'` `button_reply`/`list_reply` we already handle. Today those fall through to the `'text'` fallback with no `interactiveReplyId`, so a flow waiting on a button tap never advances (`route.ts:812`) and the inbox bubble can't render the reply chip. Mapping it to `content_type: 'interactive'` is correct — migration 010 widened the CHECK to allow it.

**That fix on its own would land today.**

### What blocks the rest

**1. `parseInt` on Meta media ids is a correctness regression.**

```js
mediaPayload = { id: parseInt(id, 10) }
```

Media ids are opaque identifiers that happen to look numeric, and they routinely run past `Number.MAX_SAFE_INTEGER` (9007199254740991 — 16 digits):

```
1013859600285441     (16 digits) -> 1013859600285441   ok
24259628093790368    (17 digits) -> 24259628093790370  LOSSY
```

Any 17-digit id silently sends the wrong number to Meta. The Cloud API takes the id as a string, so there's nothing to gain here — and it drags `string | number` through `MetaSendParameter` for all three media types.

**2. A deleted test with no replacement.**

```diff
-  it('throws an actionable error when META_APP_ID is unset', async () => {
```

Nothing in this PR touches `template-header-handle.ts`, so that test still passes on `main`. `ensureImageHeaderHandle` is the function that carried GHSA-6fr5-pxw7-r3xw, so its coverage is one I'd rather grow than shrink. If it was failing for you locally, that's worth reporting on its own — it'd be an env-leak between tests, not a reason to drop the assertion. (It also leaves a doubled blank line behind.)

**3. A 700-line UI feature is riding along under a bugfix title.**

`template-picker.tsx` is +702/-180: a tabbed rework (`map` / `media` / `advance`), a static-URL-vs-upload source switch, direct Supabase Storage uploads with MP4/PDF validation, per-type preview. That's a feature, and per CONTRIBUTING.md features are fork territory — but more immediately it's not reviewable as part of a PR described as "resolve issues in template-sending paths."

**4. Minor:** `(templateMessageParams as Record<string, unknown>)?.headerMediaUrl as string` in `send-message.ts` is an unchecked double cast. `SendTimeParams.headerMediaUrl` is already typed — the cast is only needed because `templateMessageParams` is declared `unknown` at `send-message.ts:83`. Narrowing that parameter properly is the fix. Also a stray indent on `buildHeaderComponent`'s return type.

### What I'd take

A PR containing only the `type: 'button'` webhook case — the `WhatsAppMessage.button` field, the `messageType` branch, the `parseMessageContent` case — plus a test asserting a button-tap webhook produces `content_type: 'interactive'` and a populated `interactive_reply_id`. `src/app/api/whatsapp/webhook/route.test.ts` landed on `main` today in #449 and has the harness set up for exactly that.

Keep `headerMediaId` as a string if you resend the builder change.

PR #408, opened 2026-07-19
feat: reconcile AI inbox agent with current main

davibritovilela52-design:
## Summary
- add account-scoped AI inbox agent configuration and run APIs
- add inbox/settings UI for AI agent controls
- add run guardrails, async observability, migrations, and tests

## Validation
- npm test
- npm run typecheck
- npm run build

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -63,6 +63,22 @@ const SECURITY_HEADERS = [
   },
 ] as const;
 
+export const PRIVATE_PAGE_CACHE_CONTROL =
+  "private, no-cache, no-store, max-age=0, must-revalidate";
+
+const PROTECTED_PAGE_PREFIXES = [
+  "/agents",
+  "/dashboard",
+  "/inbox",
+  "/contacts",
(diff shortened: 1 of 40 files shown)


ArnasDon:

The diff is 40 files / ~5.5k added lines, and its first commit is `feat: remove AI reply assistant, apply Decizyon rebrand`. Merging it would rename the project to Decizyon CRM, add `public/brand/*decizyon*` assets, set `DEFAULT_THEME = "decizyon"`, remove the AI reply assistant, and bring in the rest of your downstream history along with it. All of that is perfectly reasonable for Decizyon — it just isn't something upstream can take.

That said: the AI inbox agent reconciliation sounds like the most interesting piece, and if you think the template should have it, I'd gladly look at it as a focused PR against current `main`, with the rebrand and the unrelated changes left out. Opening an issue first is even better — cheaper for you than rebasing a 40-file branch only to find we disagree on scope.

Good luck with Decizyon.

PR #410, opened 2026-07-19
Worktree ai assistant

Martinxmaina:
## Test plan

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -91,3 +91,15 @@ NEXT_PUBLIC_SITE_URL=https://crm.example.com
 # Set this in CI and local development so you can exercise the full
 # template UI without a real WABA. Leave unset (or "false") in prod.
 # WHATSAPP_TEMPLATES_DRY_RUN=true
+
+# AI assistant (Settings → AI Assistant). Anthropic API key used to
+# auto-reply to inbound WhatsApp messages grounded in your knowledge
+# base. Leave blank to keep the assistant effectively off — without it
+# every inbound is handed straight to a human (the feature fails safe).
+# Get one at https://console.anthropic.com → API Keys.
+# ANTHROPIC_API_KEY=sk-ant-...
+
+# Optional override of the default model for AI replies. Defaults to
+# claude-sonnet-4-6; set claude-haiku-4-5 for a cheaper/faster option.
+# The model is also selectable per-account in Settings → AI Assistant.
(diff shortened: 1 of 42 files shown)


ArnasDon:
Closing this one. It's a large feature build, it conflicts with `main` in 18 files, and its migrations would corrupt the schema sequence for anyone who applied them.

### The migration collision is the hard blocker

The PR adds:

```
supabase/migrations/027_ai_assistant.sql
supabase/migrations/028_account_branding.sql
```

Both numbers are already taken upstream, by **different** migrations:

```
027_notifications.sql
028_webhook_endpoints.sql
```

So after a merge there'd be two `027`s and two `028`s doing unrelated things. Anyone applying migrations in order gets a schema that depends on filesystem sort order — and forks that already ran the real `027`/`028` would silently skip or double-apply depending on how they track state. This isn't a rename-and-go fix either; `027_ai_assistant.sql` is 214 lines and would need renumbering plus a re-check against everything that landed in `029`–`037` since.

### Scope

Beyond that, this is an AI assistant product: `lib/ai/` gains `anthropic.ts`, `reply.ts` (+411), `guardrails.ts`, `escalate.ts`, `prompt.ts`, `decide.ts`, `knowledge-base.ts`, `send.ts`; the UI gains `ai-assistant-settings.tsx` (+666), `knowledge-base-manager.tsx` (+627) and `branding-settings.tsx`; plus four new API routes and a 378-line design spec under `docs/superpowers/specs/`.

Worth knowing: **upstream already has an AI reply assistant** — `src/lib/ai/` with `auto-reply.ts`, `knowledge.ts`, `generate.ts`, provider adapters, and migrations `029_ai_reply`, `030_ai_knowledge`, `031_ai_reply_slot_grant`, `033_ai_reply_polish`. This would be a second, parallel implementation rather than an addition, which is most of where the 18 conflicts come from.

The test coverage is genuinely good — `reply.test.ts` (+574), `prompt.test.ts` (+319), `guardrails.test.ts` (+155). If there's a specific behaviour in your version that upstream's assistant gets wrong — a guardrail it's missing, an escalation path it doesn't handle — that framed as a focused PR against the existing `lib/ai/` would be much more likely to land than a replacement.

Also flagging: `.github/workflows/ci.yml` picks up +5 lines here, and `package.json` adds a dependency. Both worth keeping out of future upstream diffs.

PR #421, opened 2026-07-20
feat: integrate Groq API and resolve next-intl parsing issues

Arifuzzamanjoy:
Description: Hey! This PR tackles two things:

Groq API Support: Added Groq as an official AI provider so users can bring their own Groq keys (which are insanely fast). Hooked it up using their OpenAI compatibility layer, set the default model to llama-3.1-8b-instant, and ran a quick DB migration to accept 'groq' in the ai_configs table.
Translation Crashes: Fixed the INVALID_MESSAGE runtime errors that were causing the Settings pages to crash. next-intl was choking on raw HTML tags and literal curly braces {{1}} in a few of the translation strings. I swapped out t() for t.raw() in the affected components to bypass the ICU parser, and cleaned up an invalid HTML attribute in the JSON files.
Everything is tested, and the test suite is fully green (645/645 passing).

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -359,9 +359,9 @@
         "custom": "Custom Fields",
         "deals": "Deals",
         "tags": "Tags",
-      "noTags": "No tags",
-      "noDeals": "No deals",
-      "addNotePlaceholder": "Add a note..."
+        "noTags": "No tags",
+        "noDeals": "No deals",
+        "addNotePlaceholder": "Add a note..."
       },
       "actions": {
         "edit": "Edit",
@@ -932,6 +932,7 @@
         "no": "No"
       },
       "addStep": "Add step",
+      "delete": "Delete",
       "config": {
         "messageText": "Message text",
         "placeholderMessageText": "Hi! Thanks for reaching out…",
@@ -1331,13 +1332,13 @@
       "deleteLocallyAria": "Delete template locally",
       "deleteMetaLocallyTitle": "Delete from Meta and locally",
       "deleteLocallyTitle": "Delete locally",
-      
+
       "dialogEditTitle": "Edit Message Template",
       "dialogNewTitle": "New Message Template",
       "dialogEditDesc": "Save your changes to re-submit to Meta. Status will flip back to PENDING during review.",
       "dialogNewDesc": "Build a template and submit it to Meta for approval. Once approved, you can use it in broadcasts and the inbox.",
       "authWarning": "AUTHENTICATION templates have a fixed body + OTP button shape that needs a different builder. Create them in Meta WhatsApp Manager for now and use <bold>Sync from Meta</bold> to bring them in.",
-      
+
       "templateName": "Template Name",
       "namePlaceholder": "e.g. order_confirmation",
       "nameFixed": "Name is fixed once a template exists on Meta — create a new template to change it.",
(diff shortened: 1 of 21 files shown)


ArnasDon:
Closing the PR, but the second half of it found a real bug and I'm landing a fix for it — credit below. Taking the two halves separately.

## The next-intl finding is real, and worse than you described

I reproduced it against current `main`. `t()` on a message ICU can't parse doesn't crash — next-intl catches it, reports `INVALID_MESSAGE`, and renders **the key path itself**:

```
Settings.templates.bodyPlaceholder
  onError -> INVALID_MESSAGE
  t()     = "Settings.templates.bodyPlaceholder"
  t.raw() = "Hello {{1}}, your order {{2}} is confirmed."
```

So users are literally seeing `Settings.templates.bodyPlaceholder` in the template body field. Your `t.raw()` diagnosis is exactly right.

I audited all 25 strings in `en.json` containing `{{` or HTML tags and checked each against its call site. **12 are broken**, and this PR fixes 7 of them:

| file | broken keys | in this PR |
|---|---|---|
| `template-manager.tsx` | `bodyPlaceholder`, `bodyHint`, `headerTextPlaceholder`, `headerSamplePlaceholder`, `urlPlaceholder`, `urlSamplePlaceholder` | 2 of 6 |
| `whatsapp-config.tsx` | `step3_2`, `step3_3`, `step3_4`, `step4_3`, `step4_4` | 5 of 5 |
| `automation-builder.tsx` | `config.placeholderValue` | no |

The other 13 (`authWarning`, `langHint`, `scopesHint`, `noPendingDesc`, …) already use `t.rich()` with tag handlers and are fine — worth knowing so nobody "fixes" those into `t.raw()` later.

I'm opening a PR with all 12 and a regression test that fails if a new ICU-hostile string gets wired to plain `t()`. You'll be credited on it.

Two of your other i18n fixes are already on `main`: the `scopesHint` `<code className>` attribute and `useTranslations('roles')` → `'Settings.roles'`.

## Why the PR itself can't land

**1. Groq is a new feature.** Per CONTRIBUTING.md, new providers are fork territory — the template stays deliberately narrow at OpenAI + Anthropic. The implementation looks fine (OpenAI-compat layer, sensible `llama-3.1-8b-instant` default); it's a scope call, not a quality one.

**2. It rewrites 7 already-applied migrations.** 29 `uuid_generate_v4()` → `gen_random_uuid()` substitutions across `001`, `006`, `009`, `010`, `017`, `027`, `035`. Applied migrations are immutable here — every fork with a live database has already run these, so editing them changes nothing for existing installs while silently giving *new* installs a different schema history. That divergence is exactly what you can't debug later. It also doesn't achieve the usual goal of the swap, since `CREATE EXTENSION IF NOT EXISTS "uuid-ossp"` stays at the top of `001`. If you want the modernisation, it's a new forward migration with `ALTER COLUMN … SET DEFAULT`.

**3. `037` is taken.** `037_webhook_broadcast_reliability.sql` landed today in #449.

**4.** `messages/en.json` picks up unrelated whitespace churn (trailing spaces on blank lines, an indent fix) that inflates the diff.

Thanks for chasing the `INVALID_MESSAGE` errors down to the ICU parser — that was the non-obvious part, and it had been shipping broken for a while.

Issue #288, opened 2026-06-23
[bug] 502 Bad Gateway Error


pravin-justtry:
### What happened?

- login to the site, started to work afterwards moved to some other works.
- After returning to the site it does not reloaded properly and 502 bad gateway error occurs, once i remove the cookie (or) cache from browser it reloads properly.
- it happening routinely whenever we tried to work on the site after some time

### Steps to reproduce

- Frontend is does not moved to next session it tried to connect with last session but it could connect so error occurs 
- once remove the browser cookie (or) cache it connect to new session.

### What did you expect?

- Our expectation is whenever we hit the url it has to reload properly and site has to return without any hurdles.

### Where is it running?

None

### Logs / screenshots

```text

```

PR #422, opened 2026-07-22
fix: harden app security defaults

reyrodri1:
## Test plan


ArnasDon:
Closing — most of this has been overtaken by `main`, the rest is bundled in a way I can't land as one unit. But the CSP piece is right and I'd take it back scoped.

### Already landed upstream

Every dependency change here is now behind `main`:

| this PR | `main` today |
|---|---|
| `next` / `eslint-config-next` → 16.2.11 | **16.2.12** |
| `fast-uri` ^3.1.4 | **^3.1.5** |
| `hono` ^4.12.31 | **^4.12.34** |
| `ip-address` ^10.1.1 | **^10.3.1** |
| `sharp` ^0.35.3 override | already there (#463) |

`main` also carries the two `brace-expansion` overrides this predates (#470). That's where the `package.json` / `package-lock.json` / `mcp-server/package-lock.json` conflicts come from — re-resolving them would move three of those pins *backwards*.

The `useTranslations('roles')` → `'Settings.roles'` fix in `settings-overview.tsx` is also already on `main`.

### The part worth keeping

Flipping `Content-Security-Policy-Report-Only` to enforced is the documented next step — it's literally what the comment above `SECURITY_HEADERS` says to do. Scoping `'unsafe-eval'` to dev only is right, and adding `object-src 'none'` / `frame-src 'none'` / `manifest-src 'self'` costs nothing. Good catch on `worker-src 'self' blob:` too — `opus-recorder` in the inbox composer would have tripped enforcement without it.

The `private, no-store` headers on the dashboard routes are reasonable belt-and-braces as well, though I'd want the `PROTECTED_ROUTES` list checked against `src/app/(dashboard)/` so it can't silently drift when a route is added.

### What blocks it as-is

**Prettier reformat.** `next.config.ts` and `settings-overview.tsx` are requoted double → single throughout and re-wrapped. `main` isn't prettier-clean and `format:check` isn't in CI on purpose — files here follow their own local style. It turns a ~20-line security change into a 100-line diff where the real edits are hard to pick out. Please match the file's existing style rather than running the formatter.

**`src/middleware.ts` → `src/proxy.ts`.** A reasonable Next 16 migration, but it's orthogonal to security hardening and it's a rename of the file that holds all the auth redirects and the session-cookie refresh fix from #288. It deserves its own PR with its own verification, not a ride-along in a 10-file diff.

### What I'd take

One PR against current `main` touching only `next.config.ts`: the CSP key flip, the dev-only `unsafe-eval`, the four new directives, and the dashboard cache headers. No dependency changes, no rename, no reformat. Worth a note in the body on which routes you exercised with enforcement on — the composer, the media viewer, and the flow canvas are the three most likely to trip it.

PR #429, opened 2026-07-24
feat(automations): add floating AI copilot chat

davibritovilela52-design:
## Summary
- Adds a permission-gated floating AI copilot chat on `/automations`.
- Persists chat history locally in the page, renders draft automation cards, and sends serialized `{role,text}` history to the existing generation API.
- Adds English and Korean copy plus the shared popover primitive needed by the floating widget.

## Test Plan
- [x] npm test
- [x] npm run typecheck
- [x] npm run build
- [x] changed-file ESLint and git diff --check

## Notes
- Browser visual QA was not completed because the in-app browser runtime failed to initialize locally.


ArnasDon:

The diff is 152 files / ~12.7k added lines, and its first commit is `feat: remove AI reply assistant, apply Decizyon rebrand`. Merging it would rename the project to Decizyon CRM, add `public/brand/*decizyon*` assets, set `DEFAULT_THEME = "decizyon"`, remove the AI reply assistant, and bring in the rest of your downstream history along with it. All of that is perfectly reasonable for Decizyon — it just isn't something upstream can take.

That said: the floating AI copilot chat sounds like the most interesting piece, and if you think the template should have it, I'd gladly look at it as a focused PR against current `main`, with the rebrand and the unrelated changes left out. Opening an issue first is even better — cheaper for you than rebasing a 152-file branch only to find we disagree on scope.

Good luck with Decizyon.

PR #431, opened 2026-07-25
Feat/ads attribution

lrojascode:
> Fork-internal PR: `lrojascode/wacrm` `feat/ads-attribution` → `main`. Not aimed upstream — this is the "new features belong in a fork" case the template describes.

## Summary

Turns the CRM into a marketing-measurement tool: it records which ad each WhatsApp lead came from, pulls spend from Meta Ads, reports cost per lead and ROI per campaign, and tracks which salesperson owns each deal.

## What changed

**Attribution** — Meta attaches a `referral` object naming the ad on the first message after a Click-to-WhatsApp tap, and the webhook was discarding it. It is now captured (`src/lib/attribution/capture.ts`) as first-touch on the contact plus an append-only `attribution_events` row, deduped on the message id. An organic post click is recorded as `organic`, not `meta_ads` — counting it as paid would divide real spend by inflated leads and understate cost per lead.

**Meta Ads** — Graph v25.0 client (`src/lib/ads/meta.ts`) and a three-pass sync (`src/lib/ads/sync.ts`): campaigns, daily insights, and resolving captured ad ids to their campaign. Token verified against Meta before anything is stored, encrypted with the existing AES-GCM helper. `POST` for the Settings button, `GET` behind the existing cron secret for n8n.

**Google Ads / other** — no usable API (Google gates it behind a developer token they approve over weeks), so: manual campaigns with **dated** spend entries, plus tracked links (`/l/<slug>` → `wa.me` with a short opaque code the webhook reads back). Same cost-per-lead and ROI maths as Meta.

**ROI** — `deals.closed_at` (migration 039), because `updated_at` moves on any unrelated edit and would shift a deal's revenue between periods. ROI/ROAS are `null`, not `0` or `Infinity`, when spend is 0; when the ad account's currency differs from the account's, the page shows a warning instead of a number, since there is no FX conversion anywhere in this app.

**Salespeople** — board filter, per-vendor stats table (deliberately *not* filtered by that selector — it would hide the comparison it exists for), a `deal_assigned` notification, load-based round-robin, and an `assign_deal` automation step.

**Three defects found by reviewing this work, fixed in `e8aaa05`.** All three only misbehave once a day has passed, which is why same-day local testing missed them:

- Manual spend was "cumulative to date" in one row dated today, but editing it on a later day wrote a *second* row while the page sums the range — 300 corrected to 400 read as 700, and the dialog prefilled that total so each edit compounded. Now dated entries, and the dialog no longer receives the aggregate.
- A failed ad resolution was never retried (any `ad_entities` row meant "skip", and failures wrote one), so a transient Graph error or the routine "campaign not synced yet" stranded a lead's attribution permanently. Retries now widen 0/15m/1h/6h/24h and give up at 5, per `src/lib/ads/resolution-policy.ts`.
- Dates were written with the server's day (UTC in production) and read with the browser's, so spend entered late evening in Peru landed on tomorrow and vanished from a range ending "today". See `src/lib/ads/day-key.ts` for the three clocks involved and which owns what.

**Also** — follow Meta's paging (`limit=500` truncated silently, and less spend read means *inflated* ROI with nothing on screen to say so), timeout on Meta requests, batched sync upserts (~150 sequential round trips per run down to 2), and in-app numbered setup guides for Meta and Google in Settings → Ad accounts.

**Two pre-existing bugs fixed in passing** (both on `main`, both visible on screen): the role badge rendered `roles.owner` from a wrong i18n namespace, and every `Select` showed the raw value instead of the label because Base UI's `Select.Value` needs `items` on the root — derived once in `src/components/ui/select.tsx` rather than at 11 call sites.

**Migrations 037–042.** Additive, idempotent, and bundled for hand-application at `docs/deploy/ads-attribution.sql` — this self-hosted setup has no migration runner. **Apply before deploying the code**: additive changes leave the running app working, whereas new code against the old schema would not.

## Test plan

- [x] `npm run lint` — 0 errors, 39 warnings, all pre-existing (no new ones in the added files).
- [x] Feature / fix manually exercised in the browser.

**Automated:** 738 of 740 tests pass. The 2 failures are `date-utils.test.ts` (timezone-dependent) and fail on `main` too — checked by checking `main` out and running them there. New coverage: attribution capture and first-touch, ref-token round-trip, tracked links, cost-per-lead and ROI maths, round-robin, the retry policy, day-key parsing, and Meta paging/timeout.

**Migrations:** `supabase db reset` applies 001→042 cleanly from scratch — the same sequence that is then applied by hand in production. The bundled file was run twice against an already-migrated database: no errors, no data touched.

**Manually verified end to end** against a local Supabase (Docker) with two team members, using a dev script that posts the exact signed webhook payload Meta sends, so the ad path is exercised without spending ad budget:

- A Click-to-WhatsApp lead lands with source **Meta Ads** and its ad id; a repeat delivery neither duplicates nor overwrites it; `source_type: post` records as organic; a plain message stays unclassified.
- Manual Google campaign → tracked link → real 302 to `wa.me` with the code appended and the click counter incremented → inbound message carrying that code attributed to the campaign → won deal → **ROI 11%** = (500 − 450) / 450.
- The spend regression specifically: 300 dated yesterday, then **three** saves for today (400, 400, 450) leave **one** row per date holding 450, total 750. The amount field never prefills. Deleting an entry lowers the total.
- Settings: "Activate Google Ads" registers the platform, both setup accordions expand and read correctly in Spanish, and the console is clean on Campaigns, Settings, Pipelines, Automations and Flows (it previously logged a `MISSING_MESSAGE` on every Settings visit).

**Two things I could not verify, stated plainly:**

1. The retry policy against the **real** Graph API — no valid Meta token available. It is covered by 12 unit tests, and the new query was run against the real schema to rule out a bad column name, but the full loop is only confirmed once a Meta account is connected.
2. Whether the Supabase SQL Editor accepts the bundle in one paste. It rejected an earlier version because its client-side statement splitter tracks quotes without stripping comments, so an apostrophe in a comment (`the ad's URL`) desynced it — the generator now rewrites those, and the file documents running one section at a time as a fallback.

## Related

No tracking issue — this branch came out of a direct request (measure ad spend, cost per lead, ROI, and who owns each sale) rather than a filed report. The plan and the deploy runbook live in `docs/pruebas-local.md`.

## Notes for the reviewer

- No new runtime dependencies.
- Behaviour change for forkers: a new `campaigns` route and sidebar entry, five new tables, and `deals.closed_at`. Documented in `docs/pruebas-local.md`.
- The Meta Ads token needs `ads_read` only. A token from the Graph API Explorer expires within hours — production needs a Business Manager **system user** token, which is the single most likely cause of "the sync worked yesterday".


lrojascode:
error

PR #432, opened 2026-07-27
fix: setting roles translations

eldon922:
## Test plan

### DIFF
--- a/src/components/settings/settings-overview.tsx
+++ b/src/components/settings/settings-overview.tsx
@@ -40,7 +40,7 @@ export function SettingsOverview({
     useAuth();
   const { mode, theme } = useTheme();
   const t = useTranslations('Settings.overview');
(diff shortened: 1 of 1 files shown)


ArnasDon:
Closing — this one has already landed. `settings-overview.tsx:43` on `main` now reads `useTranslations('Settings.roles')`, so the diff here is a no-op against current `main`.

The same fix came in via #417 (commit 18b1b29), merged a few minutes ago. That PR was opened a week earlier and bundled two related bugs from #375 alongside this line — the `scopesHint` rich-tag and the `<SelectValue />` raw-enum rendering.

Nothing wrong with the diagnosis, though: you landed on exactly the right line and the right namespace, independently. Thanks for taking the time to file it, and sorry it turned out to be a duplicate — that's on the queue being backed up, not on you.

PR #438, opened 2026-07-27
Evolve flows module runtime and authoring

davibritovilela52-design:
## Summary
- Adds the canonical flows foundation: node registry, immutable published versions, typed runtime nodes, retries, execution records, debug inspector, approvals, analytics, flow-as-code import/export, and AI flow generation API.
- Adds first-class v2 trigger entry nodes plus durable manual/webhook/time ingress with schedule and invocation draining.
- Exposes webhook endpoint secret management and time/webhook trigger authoring support.

## Test Plan
- npm.cmd test
- npm.cmd run typecheck

Note: package-lock.json remains locally modified but was intentionally not committed.


ArnasDon:

The diff is 437 files / ~76k added lines, and its first commit is `feat: remove AI reply assistant, apply Decizyon rebrand`. Merging it would rename the project to Decizyon CRM, add `public/brand/*decizyon*` assets, set `DEFAULT_THEME = "decizyon"`, remove the AI reply assistant, and bring in the rest of your downstream history along with it. All of that is perfectly reasonable for Decizyon — it just isn't something upstream can take.

That said: the flows runtime and authoring work sounds like the most interesting piece, and if you think the template should have it, I'd gladly look at it as a focused PR against current `main`, with the rebrand and the unrelated changes left out. Opening an issue first is even better — cheaper for you than rebasing a 437-file branch only to find we disagree on scope.

Good luck with Decizyon.

PR #473, opened 2026-08-05
Integração Wacrm e Chatwoot

alexlemosrj:
Integração Wacrm e Chatwoot

## Test plan

### DIFF
--- a/.env.local.example
+++ b/.env.local.example
@@ -45,6 +45,13 @@ NEXT_PUBLIC_SITE_URL=https://crm.example.com
 # Default language locale (e.g. en)
 NEXT_PUBLIC_APP_LOCALE=en
 
+# Chatwoot Dashboard App integration. This must be the exact public origin
+# of the self-hosted Chatwoot installation, with no trailing slash. It is
+# used both to validate browser postMessage events and to allow only that
+# installation to frame /embedded.
+NEXT_PUBLIC_CHATWOOT_ORIGIN=https://chat.example.com
+CHATWOOT_DASHBOARD_ORIGIN=https://chat.example.com
+
 # ============================================================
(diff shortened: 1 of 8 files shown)


ArnasDon:
Obrigado pelo trabalho — this is a genuinely tidy PR, 8 files and no sprawl, which I appreciate. Closing it on scope rather than quality.

A Chatwoot integration is a third-party product integration, and CONTRIBUTING.md points those at forks: the template's scope is deliberately narrow, and every forker would inherit `chatwoot_deal_links`, an `/embedded` route and a `CHATWOOT_DASHBOARD_ORIGIN` env var whether or not they run Chatwoot. This is the archetypal "great idea for *your* CRM."

Some things I liked, for whoever picks this up in a fork:

- Scoping the framing relaxation to `/embedded/:path*` and leaving `X-Frame-Options: DENY` everywhere else is the right shape. Defaulting `frame-ancestors` to `'none'` when `CHATWOOT_DASHBOARD_ORIGIN` is unset is a safe fallback — easy to get backwards.
- Adding `/embedded` to `protectedPaths` in the middleware is the easy step to forget.

Two notes if you carry it forward:

1. `037` is taken — `037_webhook_broadcast_reliability.sql` landed today in #449. Renumber to `038`.
2. `chatwoot_deal_links` has no RLS policies. Every other account-scoped table here enables RLS and gates on `is_account_member(account_id)` (see `017_account_sharing.sql`). Without that, the `account_id` column is decorative and any authenticated user can read every account's links. Worth fixing in your fork.

The negative-lookahead route change (`/:path((?!embedded).*)`) is also worth a second look — it excludes any path *containing* "embedded", not just the prefix.

PR #475, opened 2026-08-06
Feature/whatsapp 24h window

Dev-Ruco:
## Test plan

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8708,17 +8708,6 @@
         }
       }
     },
-    "node_modules/next-intl/node_modules/@swc/helpers": {
-      "version": "0.5.23",
-      "resolved": "https://registry.npmjs.org/@swc/helpers/-/helpers-0.5.23.tgz",
-      "integrity": "sha512-5lSsMOTXURePglDfvuAQUqkGek9Hg2kksOYay2m0+XR++b2NWYL/4sWyuvVBIs8oKnJaxkdi9whaL/sqN13afw==",
-      "license": "Apache-2.0",
-      "optional": true,
-      "peer": true,
-      "dependencies": {
-        "tslib": "^2.8.0"
-      }
-    },
     "node_modules/node-addon-api": {
       "version": "7.1.1",
       "resolved": "https://registry.npmjs.org/node-addon-api/-/node-addon-api-7.1.1.tgz",
--- a/src/app/(auth)/signup/page.tsx
+++ b/src/app/(auth)/signup/page.tsx
@@ -73,6 +73,7 @@ function SignupPageInner() {
       password,
       options: {
         data: {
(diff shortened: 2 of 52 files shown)


ArnasDon:
Closing this — the title says "WhatsApp 24h window," but the bulk of the change is a migration of the whole app onto a private `wacrm` Postgres schema, and merging it would take every upstream install down.

### The blocker

The PR repoints every Supabase client at a schema named `wacrm`:

- `src/lib/supabase/types.ts` (new) hardcodes `SupabaseClient<any, 'wacrm', 'wacrm', any, any>`
- `middleware.ts`, `lib/supabase/client.ts` and `server.ts` gain `db: { schema: 'wacrm' }`
- 17 realtime subscriptions flip `schema: "public"` → `schema: "wacrm"`
- `signup/page.tsx` starts stamping `app: 'wacrm'` into user metadata

But **no migration in this repo creates that schema**, and this PR doesn't add one — `git diff origin/main...` touches nothing under `supabase/`. Every one of the 37 migrations builds tables in `public`. So on any install that isn't yours, every query, every realtime channel and every auth-scoped read would resolve against a schema that doesn't exist.

That's a fork-local deployment choice, which is exactly what CONTRIBUTING.md means by a stack change. Perfectly reasonable in your fork; it can't come upstream.

### The 24h window itself is good work

`getCustomerServiceWindow` is clean, and gating non-template sends on it in both `/api/whatsapp/send` and `/api/v1/messages` is the right pair of call sites. A 409 with an actionable message beats letting Meta reject the send. I'd look seriously at that as a standalone PR — it'd need:

- the `WacrmSupabaseClient` dependency dropped (take the existing client type)
- a decision on automations and flows, which today can send free-form well past 24h and would start getting 409s
- tests: window open, window closed, no inbound message ever, and template-passes-through

### Also in the way

**116 deleted comment lines.** 46% of everything this PR removes is comments — 40 lines from `api/v1/messages/route.ts` (including the whole endpoint contract block), 38 from `whatsapp/send/route.ts`, 27 from `ai/config/route.ts`. The role checks themselves survive, so this isn't a security regression, but one of the casualties is the note explaining *why* `requireRole('agent')` is on the send route:

> RLS did block the message INSERT, but the send core calls Meta BEFORE it persists, so a viewer's request still delivered a real WhatsApp message to the customer and merely failed to record it. RLS can't un-send that, so the role check belongs here.

That comment is the only record of why the check can't be left to RLS. Losing it invites someone to "simplify" the check away later.

**`package-lock.json`** drops the `next-intl/node_modules/@swc/helpers` entry. That specific removal is a known CI-breaker here — please regenerate with npm 10 rather than hand-editing.

If you want to bring the 24h window back on its own, against current `main`, with no schema changes — I'll review it.

PR #477, opened 2026-08-08
fix(i18n): read ICU-unparseable strings with t.raw() instead of t()

ArnasDon:
## Summary

Twelve UI strings have been rendering as their own keypath — the template body field literally showed `Settings.templates.bodyPlaceholder`, and the Meta setup checklist showed five keypaths where its instructions should be.

## What changed

Some catalogue strings are deliberately not ICU messages: template placeholders show the user WhatsApp's literal `{{1}}` syntax, and the setup steps carry raw `<strong class="…">` markup for `dangerouslySetInnerHTML`. next-intl's parser rejects both — and does it *quietly*. `t()` reports `INVALID_MESSAGE` to `onError` and falls back to rendering the keypath. Nothing throws, no build fails, and dev looks identical to prod:

```
Settings.templates.bodyPlaceholder
  onError -> INVALID_MESSAGE
  t()     = "Settings.templates.bodyPlaceholder"
  t.raw() = "Hello {{1}}, your order {{2}} is confirmed."
```

All twelve call sites move to `t.raw()`:

| file | n | keys |
|---|---|---|
| `template-manager.tsx` | 6 | `bodyPlaceholder`, `bodyHint`, `headerTextPlaceholder`, `headerSamplePlaceholder`, `urlPlaceholder`, `urlSamplePlaceholder` |
| `whatsapp-config.tsx` | 5 | `step3_2`, `step3_3`, `step3_4`, `step4_3`, `step4_4` |
| `automation-builder.tsx` | 1 | `config.placeholderValue` |

I audited all 25 strings in `en.json` containing `{{` or a tag. The other 13 (`authWarning`, `langHint`, `scopesHint`, `noPendingDesc`, …) already use `t.rich()` with tag handlers and are correct — no change.

`src/i18n/icu-safety.test.ts` keeps it that way: it walks `en.json`, collects every value next-intl reports `INVALID_MESSAGE` for, and fails if one is consumed by a plain `t()` in a component that opens its namespace. Filtering on `INVALID_MESSAGE` specifically — rather than any `onError` — is what keeps the well-formed `t.rich()` and interpolation sites out of the result.

## Test plan

- [x] `npm run lint` — 0 errors, 38 warnings (unchanged from `main`).
- [x] `npm test` — 73 files, 713 passed.
- [x] New test **fails with 12 offenders** when the fixes are reverted, passes with them applied — verified both directions rather than assuming.
- [ ] Not exercised in a browser: both surfaces sit behind Supabase auth, and the assertion lives at the next-intl layer where the bug actually is.

## Related

Reported by @Arifuzzamanjoy in #421, which found 7 of the 12 alongside an unrelated Groq provider. That PR is closed for scope; this lands the correctness half in full.

### DIFF
--- a/src/components/automations/automation-builder.tsx
+++ b/src/components/automations/automation-builder.tsx
@@ -1376,7 +1376,7 @@ function StepEditor({
             <Input
               value={(cfg.value as string) ?? ""}
               onChange={(e) => set({ value: e.target.value })}
-              placeholder={t("config.placeholderValue")}
+              placeholder={t.raw("config.placeholderValue")}
               className="bg-muted text-foreground"
             />
           </FieldBlock>
--- a/src/components/settings/template-manager.tsx
+++ b/src/components/settings/template-manager.tsx
@@ -776,7 +776,7 @@ export function TemplateManager() {
                   <Input
                     id="template-header-text"
                     aria-label="Header text"
-                    placeholder={t('headerTextPlaceholder')}
+                    placeholder={t.raw('headerTextPlaceholder')}
                     value={form.header_content}
                     onChange={(e) =>
                       setForm({ ...form, header_content: e.target.value })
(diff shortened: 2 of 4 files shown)


Arifuzzamanjoy:
Thanks for this credit.
Will contribute more soon.

PR #484, opened 2026-08-10
Fix blank inbox bubble for template messages sent via API

### LINKED ISSUE #483: [bug] Template messages persist content_text = null and render as an empty bubble
### What happened?

A template message sent through the public API (POST /api/v1/messages with type: "template" and no "text" field) is delivered by Meta and saved to the messages table, but with content_text = null. The Inbox thread renders it as a completely empty bubble - timestamp and delivery ticks are there, no content at all. The conversation-list preview reads "[template]".

The same defect affects every automation template send, unconditionally.

### Steps to reproduce

1. Connect WhatsApp and sync an approved template that has body variables, e.g. body_text = "Your order {{1}} ships on {{2}}".
2. Create an API key with the messages:send scope.
3. POST /api/v1/messages with no "text" field:

   {
     "to": "+15551234567",
     "type": "template",
     "template": { "name": "order_update", "language": "en", "params": ["A123", "Friday"] }
   }

4. Open /inbox and select the conversation.

Observed: the outbound bubble is empty. The row confirms it:
  select content_type, content_text, template_name from messages order by created_at desc limit 1;
  -> ('template', NULL, 'order_update')

The message itself was delivered fine - status reaches 'read'.
Sending the same template from the dashboard composer renders correctly, which is what makes this easy to miss.

### What did you expect?

The stored message should carry the substituted template body - the same text the dashboard composer displays when it sends the same template - and the conversation-list preview should show it instead of "[template]".

### Commit / version

89827b5

### Where is it running?

Other Node host

### Logs / screenshots

```text
Root cause:

- sendMessageToConversation (src/lib/whatsapp/send-message.ts) persists only the caller's contentText. It loads the message_templates row a few lines earlier for the send-builder, but never uses body_text for the stored copy. The dashboard composer masks the bug by pre-rendering client-side (renderTemplateBody in src/components/inbox/message-thread.tsx) and posting the result as content_text; the public API has no equivalent.

- src/lib/automations/meta-send.ts has the same defect unconditionally:
      const content_text = input.kind === 'text' ? input.text : null
  so every automation template send stores null.

- The Inbox renders a template's body only `if (message.content_text)` (src/components/inbox/message-bubble.tsx). The fallback "Template" chip is styled bg-primary/20 + text-primary, but an outbound bubble is bg-primary + text-primary-foreground - primary-on-primary, so the chip is invisible and nothing at all renders.

Possibly related: the template lookup uses .eq('language', templateLanguage || 'en_US') while templates synced from Meta commonly carry 'en'. A caller that omits language matches no row, so the body cannot be resolved even once the above is fixed. Happy to split this into its own issue.

I have a fix ready with 12 unit tests; lint, typecheck, test and build all pass, and it is verified on a live deployment. Glad to open a PR if this looks aligned.
```

neerajgupta2407:
## Summary

Template messages persisted `content_text = null` unless the caller happened to
pass pre-rendered text, so they rendered in the Inbox as a completely empty
bubble. This renders the body server-side in the shared send core, so every
caller — public API, dashboard, automations — stores readable text.

## What changed

- **`src/lib/whatsapp/template-render.ts`** (new) — `renderTemplateBody()` does the
  `{{N}}` substitution, leaving unmatched placeholders intact rather than blanking
  them. `resolveTemplateBodyParams()` picks body values out of either shape a
  caller may use (structured `messageParams.body` or the legacy positional array),
  mirroring how `sendTemplateMessage` resolves the same pair so the persisted copy
  can't drift from what Meta was actually sent.
- **`src/lib/whatsapp/send-message.ts`** — when the message is a template and the
  caller supplied no `contentText`, derive it from the `message_templates` row the
  function already loads for the send-builder. Caller-supplied text still wins, so
  the dashboard path is unchanged. Same value feeds `last_message_text`, replacing
  the `[template]` placeholder in the conversation list.
- **`src/lib/automations/meta-send.ts`** — had the same defect unconditionally
  (`content_text = input.kind === 'text' ? input.text : null`), so *every*
  automation template send was blank. Now loads the template row and renders it.
- **`src/components/inbox/message-bubble.tsx`** — a template with no resolvable body
  now falls back to naming the template instead of rendering nothing. Also fixes
  the "Template" chip, which was `bg-primary/20 text-primary` inside an outbound
  bubble that is `bg-primary text-primary-foreground` — primary-on-primary, so it
  was invisible and the bubble appeared entirely empty. It now follows the same
  `onPrimary` pattern `ReplyQuote` already uses.
- **`src/components/inbox/message-thread.tsx`** — drops its local copy of the
  substitution helper in favour of the shared one; behaviour unchanged.
- **`messages/en.json`, `messages/ko.json`** — two keys for the fallback label.

Tests: 12 new cases — 7 for the helper (substitution, unmatched placeholder,
repeated placeholder, both param shapes) and 5 driving the send core against a
stubbed client, asserting the persisted row. Those cover the regression itself,
that caller-supplied text still wins, that a missing template row doesn't fail
the send, and that plain text sends are untouched.

## Screenshots
Before
wacrm-bug-before-redacted

After
wacrm-bug-after-redacted

Same template, sent through `POST /api/v1/messages` with no `text` field.

| Before | After |
| --- | --- |
| _drag the empty-bubble image here_ | _drag the rendered-body image here_ |

## Test plan

- [x] `npm run lint` — no new errors (0 errors; the 38 pre-existing warnings are
      unchanged).
- [x] `npm test` — 725 passing, including the 12 new cases.
- [x] Feature / fix manually exercised in the browser — verified on a live
      deployment: a template sent via `POST /api/v1/messages` with no `text`
      field now renders the substituted body in the Inbox, and the
      conversation-list preview shows it instead of `[template]`. See the
      screenshots above.

The bug was also confirmed against a real occurrence before the fix: a row with
`content_type='template'`, `template_name` set, `content_text=NULL` and
`status='read'` — Meta had delivered it — which rendered as the empty bubble on
the left. After deploying, an identical request (same template, same shape, no
`text`) persisted the fully substituted body; that message is the one on the
right. The pre-existing NULL row is untouched, since the change only affects what
is written at send time — it now falls back to naming the template rather than
rendering nothing.

## Note on formatting

I did not run `npm run format`, deliberately. It's `prettier --write .`, and five
of the files this PR touches were already non-conforming on `main`, so running it
would have buried the change under whole-file reformatting. The two files added
here are Prettier-clean; everything else is left exactly as it was. Happy to
follow up with a separate formatting-only PR if that's wanted.

## Related

```
Closes #483
```

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -212,6 +212,8 @@
       "document": "Document",
       "locationShared": "Location shared",
       "template": "Template",
+      "templateNamed": "Sent template \"{name}\"",
+      "templateNoPreview": "Template sent",
       "buttonReply": "Button reply",
       "interactiveReply": "[Interactive reply]",
       "unsupported": "[Unsupported message type]",
--- a/messages/ko.json
+++ b/messages/ko.json
@@ -212,6 +212,8 @@
       "document": "문서",
       "locationShared": "위치 공유됨",
       "template": "템플릿",
+      "templateNamed": "\"{name}\" 템플릿을 보냈습니다",
+      "templateNoPreview": "템플릿을 보냈습니다",
       "buttonReply": "버튼 응답",
       "interactiveReply": "[인터랙티브 응답]",
       "unsupported": "[지원되지 않는 메시지 형식]",
--- a/src/components/inbox/message-bubble.tsx
+++ b/src/components/inbox/message-bubble.tsx
@@ -61,10 +61,13 @@ function MessageContent({
   message,
   t,
   onOpenMedia,
+  onPrimary,
 }: {
   message: Message;
   t: ReturnType<typeof useTranslations>;
   onOpenMedia?: (messageId: string) => void;
+  /** True inside an outbound bubble, which is filled with `bg-primary`. */
+  onPrimary?: boolean;
 }) {
   // Passed to the media bubbles as a no-arg callback; `undefined` when the
   // parent wired up no viewer, which is what makes them non-clickable.
@@ -130,15 +133,32 @@ function MessageContent({
     case "template":
       return (
         <div>
-          <span className="mb-1 inline-flex items-center gap-1 rounded bg-primary/20 px-1.5 py-0.5 text-[10px] font-medium text-primary">
+          <span
+            className={cn(
+              "mb-1 inline-flex items-center gap-1 rounded px-1.5 py-0.5 text-[10px] font-medium",
+              // Outbound bubbles are `bg-primary`, so a primary-on-primary
+              // chip is invisible against them.
+              onPrimary
+                ? "bg-primary-foreground/20 text-primary-foreground"
+                : "bg-primary/20 text-primary",
+            )}
+          >
             <LayoutTemplate className="h-3 w-3" />
             {t("template")}
           </span>
-          {message.content_text && (
-            <p className="mt-1 whitespace-pre-wrap break-words text-sm">
-              {message.content_text}
-            </p>
-          )}
+          {/* A template whose body we couldn't resolve locally persists
+              no content_text. Naming it beats rendering an empty bubble. */}
+          <p
+            className={cn(
+              "mt-1 whitespace-pre-wrap break-words text-sm",
+              !message.content_text && "italic opacity-80",
+            )}
+          >
+            {message.content_text ||
+              (message.template_name
+                ? t("templateNamed", { name: message.template_name })
+                : t("templateNoPreview"))}
+          </p>
         </div>
       );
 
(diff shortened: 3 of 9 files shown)


neerajgupta2407:
Closing this — #483 (4c9c9eb) landed the same fix on main, and covers more ground than this PR did: it also routes the broadcast API and `broadcast-core.ts` through the shared resolver, and adds the `en`/`en_US` language fallback that this branch didn't handle. `template-body.ts` supersedes the `template-render.ts` helper here.

Thanks for the pointer on the invisible `bg-primary/20` chip — good to see that fixed too.

PR #498, opened 2026-08-12
ci: replay migrations against a real Postgres, replacing the dead Supabase check

ArnasDon:
Draft while I confirm the job goes green — I have no container runtime locally, so this is the first time it can actually run.

Replaces the `Supabase Preview` check, which has never validated anything. See the commit message for the diagnosis; short version: two stale dashboard integrations, one 404ing and one hanging, and Branching needs a `supabase/config.toml` this repo never had. Every migration in this repo has shipped unexecuted by CI.

This job boots a throwaway Postgres with the Supabase CLI, replays every migration from nothing, and then asserts a few of the resulting objects exist (the DDL here is all `IF NOT EXISTS`-guarded, so "no error" alone doesn't prove anything got built). No secrets, no Supabase account, no paid plan.

**You still need to disconnect the two dashboard integrations by hand** — Integrations → GitHub in the Supabase dashboard. That can't be done from the repo, and until it is, the dead check keeps appearing alongside this one.


ArnasDon:
Out of draft — the job is green and, more to the point, proven able to go red.

**What CI established (not assumed):**

- All 38 migrations on `main` apply to a clean Postgres 17 with Supabase's `auth`/`storage` schemas present. First time that has ever run.
- The check can fail. A deliberately false assertion (commit `42c7db0`, run [31579334056](https://github.com/ArnasDon/wacrm/actions/runs/31579334056)) turned the job red with a clear message, then I reverted it. Worth doing given this PR exists because the old check passed by not working.
- `supabase/ci/verify-schema.sql` must be exactly **one** statement — `db query --file` sends it as a prepared statement and a second top-level `DO` dies with "cannot insert multiple commands into a prepared statement". Found the hard way (`f91a6c8`); documented in the file.
- **Migration 039 from #496 applies cleanly** and produces the columns and widened bucket allow-list it claims (`65c99d8`). I borrowed it into a run here and then stripped it, so that PR's open "never executed anywhere" risk is now closed.

Runtime is ~1m45s, and it only fires on `supabase/**` changes.

Two caveats:

- `config.toml` sets `major_version = 17`, which is the CLI default and **not** a reading of the live project. If production is on 15, align it — otherwise this check can green-light SQL production would reject, which defeats the point.
- The two dead dashboard integrations still need removing by hand (Integrations → GitHub). Until then `Supabase Preview` keeps appearing next to this job; you can see it going `skipped` on the commits here that didn't touch `supabase/migrations/`.

PR #520, opened 2026-08-23
feat(i18n): Spanish locale with per-request detection

HaroldPater:
Adds Spanish alongside English, detected from the browser and
overridable per user.

## What changed

**Locale is resolved per request.** It used to come from
`NEXT_PUBLIC_APP_LOCALE`, baked in at build time, so a deployment
served one language to everyone. Now: an explicit `NEXT_LOCALE` cookie
wins, then the browser's `Accept-Language`, then English. `es-MX`,
`es-419` and `ES-ES` all match Spanish; quality weights are honoured
over header order. `NEXT_PUBLIC_APP_LOCALE` still sets the
deployment-wide default for the last step, but no longer overrides a
user's pick.

**Spanish catalogue.** All 1468 keys of `en.json`, with ICU
placeholders, plural forms, the literal `{{1}}` markers Meta requires,
and the inline tag names read by `t.rich()` preserved. English has no
per-key fallback, so the parity test now covers `es.json` too — a
missing key fails the build instead of rendering as its own keypath.

**Picker in Settings → Appearance.** Automatic / English / Español.
Writes the cookie and refreshes; no save button, matching the mode and
accent controls beside it.

`ko.json` stays in the repo but is not offered in the picker — add
`ko` to `SUPPORTED_LOCALES` to turn it on.

## Two consequences worth reviewing

**Pages are no longer prerendered static.** Reading the cookie and
header makes every route dynamic (`/login`, `/dashboard` etc. went
from static to server-rendered on demand). That is inherent to
detecting the language without locale-prefixed routes; the
alternative is `/[locale]/...`, which changes every URL.

**Page HTML is no longer shared-cacheable.** `next.config.ts`
advertised pages as `public, s-maxage=300` for the CDN. With a
per-user locale that lets a shared cache serve Spanish to someone who
asked for English. The usual fix — `Vary: Accept-Language, Cookie` —
does not work here: Next.js sets its own `Vary` on these routes and
overwrites both a configured header and one appended from middleware
(verified against a production build). Changed to `private,
no-store`. Hashed assets under `/_next/static` are untouched and still
cache immutably.

Note this hazard predates the change: dashboard HTML rendered against
a user's Supabase session was already advertised as publicly cacheable
without `Vary: Cookie`. The locale only made it easier to trigger.

## Verification

841 tests pass, typecheck clean, lint 0 errors (36 pre-existing
warnings). Checked against a production build and the Docker
container: a Spanish browser gets `lang="es"` and Spanish copy, a
Korean or headerless one gets English, and the cookie beats the
browser in both directions.


HaroldPater:
Opened by mistake against the wrong repository — this targets my own fork, not upstream. Closing. Apologies for the noise.

Issue #474, opened 2026-08-05
[bug] Send Buttons Issue in Automations Section


angelTDW:
### What happened?

<img width="374" height="805" alt="Image" src="https://github.com/user-attachments/assets/b1366fe3-399b-46b7-b96c-a01876c2cd9b" />

when ever i add a send button under automations section, this is how its shown, thou i can add data in it.
but, i am not able to enter any data when the send buttons is added under conditions node. i only receive errors as can be seen in console.

<img width="1464" height="734" alt="Image" src="https://github.com/user-attachments/assets/15c61054-f0a6-4db8-8df6-b980b32086eb" />

### Steps to reproduce

add any node prior,
add a conditions node
add a send buttons node under yes or no

and it doesnt work. css is wrecked as well as cant type anything.

### What did you expect?

properly displayed, and message should be added without any issues.

### Commit / version

v0.8.0

### Where is it running?

Hostinger VPS

### Logs / screenshots

```text
0nbt8d8tj5nk4.js:3 u: MISSING_MESSAGE: Automations.builder.delete (en)
    at f (0nbt8d8tj5nk4.js:3:7892)
    at p (0nbt8d8tj5nk4.js:3:8079)
    at g (0nbt8d8tj5nk4.js:3:8538)
    at et (09dlay_sh191_.js:1:22101)
    at ay (12wak83h20mr8.js:1:62886)
    at oJ (12wak83h20mr8.js:1:83143)
    at iu (12wak83h20mr8.js:1:94745)
    at sd (12wak83h20mr8.js:1:137985)
    at 12wak83h20mr8.js:1:137830
    at sf (12wak83h20mr8.js:1:137838)
en	@	0nbt8d8tj5nk4.js:3
f	@	0nbt8d8tj5nk4.js:3
p	@	0nbt8d8tj5nk4.js:3
g	@	0nbt8d8tj5nk4.js:3
et	@	09dlay_sh191_.js:1
ay	@	12wak83h20mr8.js:1
oJ	@	12wak83h20mr8.js:1
iu	@	12wak83h20mr8.js:1
sd	@	12wak83h20mr8.js:1
(anonymous)	@	12wak83h20mr8.js:1
sf	@	12wak83h20mr8.js:1
se	@	12wak83h20mr8.js:1
sQ	@	12wak83h20mr8.js:1
sU	@	12wak83h20mr8.js:1
sV	@	12wak83h20mr8.js:1
[...]

PR #528, opened 2026-08-26
feat(whatsapp): Evolution Go provider, connection reliability, and inbox UX

karlmarxmint:
## Summary

- **Evolution Go as an alternative WhatsApp provider** — self-hosted, whatsmeow-based, no Meta Business account needed. QR pairing, its own inbound webhook, admin-gated instance lifecycle, and a danger-zone conversation reset for switching numbers/providers cleanly. `send-message.ts` and the automation/flow senders branch on `whatsapp_config.provider` so messages are indistinguishable in the CRM regardless of transport.
- **Evolution Go connection reliability** — a polling health-check (`/api/whatsapp/evolution/health-check`) detects and recovers a silently-dropped whatsmeow session, backing off between reconnect attempts. Explicit read receipts (sent only when an agent opens the conversation) and inline media upload for outbound sends.
- **Message templates on Evolution Go** — no Meta approval step; templates render down to plain text for the send.
- **Inbox UX** — "send to pipeline" (create a deal from the current thread), sound/desktop notification for new messages, ordering fixes.
- **i18n** — full pt-BR locale (new), a pre-login language switcher, and ko.json parity for all of the above.
- Three unrelated fixes picked up along the way: `time_of_day` automation condition now evaluates in the account's business timezone instead of the container's; account-scoped ownership checks on automations that were previously author-scoped; the OTP-based password reset flow (superseded upstream — see note below).

## Notes for the reviewer

- This branch sat unpushed for a while, so it was rebased onto current `main` (38 commits happened here in the meantime, several overlapping the same files). Two migrations collided on number with ones already merged upstream and were renumbered: `037_whatsapp_provider.sql` → `040_whatsapp_provider.sql`, `038_evolution_health_check.sql` → `041_evolution_health_check.sql`.
- A couple of independent fixes converged on the same bugs already fixed upstream (the nested-condition-branch step path in `automation-builder.tsx`, PR #474) — the upstream version was kept, this branch's duplicate work was dropped in the rebase.
- The password-reset-link-to-OTP commit here predates upstream's own auth fixes on `main`; kept it since Evolution accounts still hit the same Gmail-link-scanner issue, but flagging in case upstream already addressed this differently.
- `npm run typecheck`, `npm run lint`, and `npm test` (835/835) all pass post-rebase.

## Test plan

- [ ] `npm run typecheck && npm run lint && npm test`
- [ ] Connect an account via Evolution Go (QR pairing) and confirm send/receive both directions
- [ ] Kill the Evolution Go session externally and confirm the health-check reconnects it
- [ ] Send a message template on an Evolution-provider account
- [ ] "Send to pipeline" from an inbox thread creates a deal
- [ ] Switch UI language pre-login via the new language switcher

https://claude.ai/code/session_01DP3Ua7KpuW7g4yZ9Gm1eAE


ArnasDon:
Thanks for the detailed write-up — this is clearly a lot of careful work. I can't merge it as one 92-file PR, though, for the reasons in `CONTRIBUTING.md`: a second WhatsApp transport (Evolution Go), a locale switcher, inbox notification sounds, send-to-pipeline and a new pt-BR catalogue are all features, and the template deliberately keeps its scope narrow so forks stay easy to rebase. Two practical problems on top of that: the branch is now `CONFLICTING` with `main`, and migrations `040_whatsapp_provider.sql` / `041_evolution_health_check.sql` collide again with `040_contact_business_scoped_user_id.sql` / `041_fix_broadcast_contact_id_ambiguity.sql` that landed since your rebase (main is at 042). pt-BR and es locales are also being added upstream as part of #375, so that part will conflict outright.

Three pieces in here **do** belong upstream and don't exist on `main` yet — I'd very much like them as separate small PRs, each against current `main` with its test:

1. **Public webhook body-size cap** — `src/lib/http/read-body-with-limit.ts` + its use in the webhook route. `main` has no cap today.
2. **`time_of_day` in the account's business timezone** — `main` still does `new Date().getHours()` in the container's zone (`src/lib/automations/engine.ts`, `case 'time_of_day'`).
3. **Account-scoped automation ownership** — `src/app/api/automations/[id]/route.ts` on `main` still filters `.eq('user_id', user.id)` on GET/PUT/DELETE, so teammates on the same account can't manage each other's automations while the rest of the app is account-scoped.

The password-reset changes I'd hold: `main` already has `forgot-password`, and the OTP variant should be discussed in an issue first so we don't end up with two flows.

Once the split PRs are up I'll suggest closing this one (or keep it going in your fork — that's exactly what the template is for). If you'd rather I cherry-pick the three commits myself and open the PRs crediting you, say so and I will.

PR #529, opened 2026-08-26
fix(contacts): surface the real reason a CSV import row failed

karlmarxmint:
## Summary

- The CSV contact import (`import-modal.tsx`) swallowed the actual Postgres/PostgREST error for any per-row insert failure that wasn't a unique violation — the user only ever saw "N contacts failed to import" with no phone, name, or underlying reason. The real error is now captured and shown per-row (name/phone + reason) in the import result panel.
- `dedupeByPhone` folded rows with an unparseable/empty phone into the same `duplicates` counter as real in-file collisions, so a contact that never duplicated anything was reported to the user as a dupe. Split into a separate `invalid` count/result line.
- `parseContactCsv` silently dropped rows with an empty phone cell before they were ever counted anywhere — the parsed row count came up short of the file with no indication why. They're now kept and land in the same `invalid` bucket.

Found while investigating a real report: a user imported a CSV where 2 valid-looking contacts were rejected with zero explanation, while an actual duplicate in the same file was accepted (that acceptance is a separate, known gap — bulk import dedupes on exact normalized-phone match only, not the trunk-prefix-tolerant fuzzy match the webhook/manual-form paths use elsewhere — not changed here to keep this fix scoped).

## Test plan

- [x] `npm run typecheck && npm run lint && npm test` (826/826 passing)
- [ ] Import a CSV with a row that fails validation server-side (e.g. RLS) and confirm the reason shows in the result panel
- [ ] Import a CSV with a blank-phone row and confirm it's reported as "invalid phone", not silently dropped or counted as a duplicate

https://claude.ai/code/session_01DP3Ua7KpuW7g4yZ9Gm1eAE

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -455,7 +455,11 @@
       "resultTags": "{count} tag assigned",
       "resultTags_plural": "{count} tags assigned",
       "resultSkipped": "{count} skipped",
+      "resultInvalidPhone": "{count} invalid phone",
+      "resultInvalidPhone_plural": "{count} invalid phones",
       "resultFailed": "{count} failed",
+      "failedRowsHeading": "Failed rows",
+      "unknownReason": "Unknown error",
       "willBeCreated": "{name} (will be created on import)",
       "columns": {
         "phone": "Phone",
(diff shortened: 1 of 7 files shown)


ArnasDon:
Reviewed. This looks good and I've verified it on current `main`: clean merge, `eslint` 0 errors, `tsc --noEmit` clean, vitest 993/993, `next build` passes.

Checked the one other consumer of the changed helpers: `parseBroadcastCsv` (`src/lib/broadcast-csv.ts`) now receives empty-phone rows from `parseContactCsv` and `dedupeByPhone` files them under the new `invalid` count, which that path ignores — so broadcast CSV behaviour is unchanged (those rows were dropped uncounted before, and still are). No action needed; noting it for the record.

Queued for merge.

Issue #472, opened 2026-08-05
[bug] BroadCast Issue


truesyssoftech:
### What happened?

If i do broadcast for 1000 Total Recipients it is stop sending after 200 to 250 , after that not further activity like no campaign status is updated or no further sending 

So seems it fails some where silently

<img width="1660" height="467" alt="Image" src="https://github.com/user-attachments/assets/edd06cf2-e56a-4922-8389-9f9661f72064" />

### Steps to reproduce

1. Sign In To Self Hosted Version
2. Add Contacts Around 1000
3. Start Campaign With Template For All Contacts (Make Sure Total Recipients will be more around 1000)
4. Process Campaign and monitor progress

### What did you expect?

1. It should process campaign without failing or update status properly if stuck any where
2. Option to reprocess pending
3. Option to reprocess failed one

### Where is it running?

Other Node host

### Logs / screenshots

```text

```

angelTDW:
did u check, your allowed limit for a day is more then 250 in meta panel?

ArnasDon:
Found it, and it's our own rate limiter — #481 has the fix. @angelTDW's question about the Meta daily cap is a fair one, but that isn't what's happening here.

The dashboard doesn't send a campaign in one request. It fans out over `/api/whatsapp/broadcast` in batches of 10 recipients with a 1s pause, so your 1,000-recipient send is ~100 separate calls over several minutes. That endpoint was rate-limited to **5 requests per minute per user** — and the comment above the limit explains why it was set that low: *"even a 1 000-recipient broadcast is one call"*. It isn't. So 5 batches (50 recipients) cleared each minute and every other batch came back 429, at which point the wizard wrote off all ten of its recipients as failed. Effective ceiling ~50 recipients/minute, everything past it discarded — which is the plateau you're seeing. If you look at the failed recipients on that campaign, they should carry `Rate limit exceeded` as the error.

#481 does two things: sizes the bucket (60/min) for the call pattern that actually exists, and makes the wizard wait out `Retry-After` and retry the batch instead of failing ten people. Only 429s are retried — the limiter rejects before we call Meta, so a 429'd batch provably sent nothing, whereas replaying after a 500 or a dropped connection could message someone twice.

[...]

Issue #370, opened 2026-07-11
[BUG] Failed broadcast recipient insertion leaves an orphaned sending broadcast


iuryeng:
## Summary

Broadcast creation persists the parent row before inserting recipients. If recipient insertion fails, the function throws but leaves the parent broadcast persisted with status `sending` and no recipients.

## Environment

- Version: `0.8.0`
- Commit: `b8677608015790194c14ab7c85943ce26c7f9ef0`
- Reproduction: Vitest characterization with synthetic data
- Repeated executions: 3

## Steps to reproduce

1. Call `createBroadcast()` with a valid synthetic broadcast and recipient list.
2. Allow the parent broadcast insert to succeed.
3. Make the recipient insert return a database error.
4. Inspect the persisted parent state after the function throws.

```powershell
npm test -- --run src/lib/whatsapp/broadcast-partial-repro.test.ts
```

No production credentials, phone numbers, customer content, or external services are used.

## Expected behavior

Broadcast creation should be atomic. On recipient failure, no `sending` parent should remain, or the parent should move to an explicit failed state with a safe retry path.

## Actual behavior

The function throws `BroadcastError`, but the parent remains:

```text
Parent broadcast: created
Status:           sending
Recipient rows:   0
Function result:  BroadcastError
```

The result was identical in three consecutive focused test runs.

## Technical evidence

The parent is inserted first:

https://github.com/ArnasDon/wacrm/blob/b8677608015790194c14ab7c85943ce26c7f9ef0/src/lib/whatsapp/broadcast-core.ts#L197-L214

Recipients are inserted in a separate operation, and failure throws without compensating for the parent:

https://github.com/ArnasDon/wacrm/blob/b8677608015790194c14ab7c85943ce26c7f9ef0/src/lib/whatsapp/broadcast-core.ts#L215-L235

## Impact

Transient database failures can leave campaigns that appear to be sending but have no delivery plan. Operators may retry and create additional orphaned campaigns, while operational views and counts contain misleading `sending` records.

## Scope and limitations

This is a route-logic/unit-integration characterization with a Supabase chain mock. It demonstrates the missing compensation boundary in the code; it does not claim a live database outage or a specific production failure rate.

## Suggested acceptance criteria

- Parent and recipient rows are created atomically, preferably through a transaction/RPC.
- A recipient insertion failure leaves no orphaned `sending` parent.
- Retry behavior is explicit and does not create duplicate campaigns.
[...]

PR #537, opened 2026-08-29
fix(broadcasts): qualify contact_id so create_broadcast_with_recipients can execute (#536)

### LINKED ISSUE #536: [bug] create_broadcast_with_recipients aborts with 42702 "contact_id is ambiguous" — POST /api/v1/broadcasts has never worked
### What happened?

`POST /api/v1/broadcasts` fails 100% of the time with a 500. The database function it calls, `create_broadcast_with_recipients`, cannot execute at all — it aborts with SQLSTATE 42702:

```
column reference "contact_id" is ambiguous
It could refer to either a PL/pgSQL variable or a table column.
```

The function is declared `RETURNS TABLE(broadcast_id UUID, recipient_id UUID, contact_id UUID)`. In PL/pgSQL a `RETURNS TABLE` output column is **also an in-scope variable**, so inside the body `contact_id` names a variable. The recipient insert then ends in a bare `RETURNING id, contact_id`, which also matches `broadcast_recipients.contact_id`. Postgres refuses to resolve it.

**Location:** `supabase/migrations/038_broadcast_resume.sql:105` (and originally `037_webhook_broadcast_reliability.sql:144`, which introduced the function).

```sql
  RETURN QUERY
  WITH ins AS (
    INSERT INTO broadcast_recipients (
      broadcast_id, contact_id, status, template_params
    )
    SELECT v_broadcast_id, t.cid, 'pending', t.prm
    FROM unnest(p_contact_ids, p_template_params) AS t(cid, prm)
    RETURNING id, contact_id          -- <-- 42702 here
  )
  SELECT v_broadcast_id, ins.id, ins.contact_id
  FROM ins;
```

### Why it went unnoticed

Two things hid it, and the second one seems worth fixing on its own:

1. **Only the public API path uses this function.** `src/lib/whatsapp/broadcast-core.ts:202` is the sole caller, reached from `POST /api/v1/broadcasts`. The dashboard's own broadcast route (`POST /api/whatsapp/broadcast`) loops `sendTemplateMessage` and never writes a campaign row, so the UI looks perfectly healthy while `broadcasts` and `broadcast_recipients` stay empty.

2. **A plpgsql body is only parsed at `CREATE` time — name resolution happens on first execution.** So the migration applies cleanly and CI goes green on a function that can never run. Any check that *applies* migrations without *calling* the RPCs they define will keep missing this class of bug.

The other identifiers in the body are fine: `broadcast_id` appears only in an INSERT column list (never a variable reference), and the final `SELECT` reads through `ins`.

### Steps to reproduce

1. Apply migrations through `038` to a clean database.
2. Create an API key with the `broadcasts:send` scope.
3. `POST /api/v1/broadcasts` with an approved template and at least one contact.
4. Response is a 500; server logs show `[broadcast-core] create broadcast error:` with the 42702 above. `broadcasts` has no new row.

Or straight from psql, no app needed:

```sql
SELECT * FROM create_broadcast_with_recipients(
  '<account_id>', '<user_id>', 'probe', '<template>', 'en_US',
  1, ARRAY['<contact_id>']::uuid[], ARRAY['[]'::jsonb]
);
-- ERROR:  column reference "contact_id" is ambiguous
```

### What did you expect?

The function creates the campaign plus its recipient rows and returns one row per recipient, as `broadcast-core.ts` expects.

### Suggested fix

One qualified identifier — this is the whole change:

```diff
-    RETURNING id, contact_id
+    RETURNING id, broadcast_recipients.contact_id
```

Qualifying it names the column and nothing else. Signature, arguments and result columns are unchanged, so `broadcast-core.ts` needs no edit. (`#variable_conflict use_column` at the top of the body would also work if you prefer that idiom — I went with the qualification since it's local to the line that's actually ambiguous.)

I've verified the fix against a live database: with the qualification in place the function creates the campaign and returns its recipient row, executed inside `BEGIN … ROLLBACK` so nothing persisted.

```
NOTICE:  PROBE OK: 1 recipient row(s) returned
```

Happy to open a PR with the corrected function if that's useful.

### Commit / version

`6ed9191` (upstream/main as of 2026-08-12) — the bare `RETURNING id, contact_id` is still present at `038_broadcast_resume.sql:105`.

Introduced in #388/#449 (migration `037`, fixing #370) and carried forward unchanged by #495 (migration `038`, fixing #472).

neerajgupta2407:
Fixes #536.

`POST /api/v1/broadcasts` has never succeeded. `create_broadcast_with_recipients` is declared `RETURNS TABLE(broadcast_id, recipient_id, contact_id)`, and in PL/pgSQL a `RETURNS TABLE` output column is **also an in-scope variable**. The recipient insert ends in a bare `RETURNING id, contact_id`, so that name resolves against both the target table's column and the function's own output variable, and Postgres aborts:

```
42702: column reference "contact_id" is ambiguous
       It could refer to either a PL/pgSQL variable or a table column.
```

## The change

One qualified identifier:

```diff
-    RETURNING id, contact_id
+    RETURNING id, broadcast_recipients.contact_id
```

Qualifying it names the column and nothing else. Signature, arguments and result columns are unchanged from `038`, so `lib/whatsapp/broadcast-core.ts` needs no edit — this is a migration-only PR.

The other identifiers in the body are already unambiguous: `broadcast_id` appears only in an INSERT column list (never a variable reference), and the final `SELECT` reads through `ins`.

## Why a new `040_` file instead of editing `038`

Applied migrations are recorded in `schema_migrations` — that ledger is what made the duplicate-version collision in `034_fix_profiles_update_rls.sql` a hard failure (23505). So editing `038` in place would fix only *fresh* installs: every existing deployment already has `038` recorded and would keep the broken function forever.

`040` is a `CREATE OR REPLACE` of the exact `038` signature, so it is idempotent and safe to re-run. This is the same shape as `034` repairing `017`'s policy. Happy to switch to an in-place edit of `038` instead if you'd rather — it's your call on how you want existing installs handled.

## Verification

Both function bodies — the current `038` one and the fixed one — created side by side in a clean Postgres 16, then called:

```
>>> BOTH FUNCTIONS CREATED WITHOUT ERROR (this is the trap)

>>> calling upstream 038 body:
ERROR:  column reference "contact_id" is ambiguous
LINE 5:     RETURNING id, contact_id
                          ^
DETAIL:  It could refer to either a PL/pgSQL variable or a table column.
CONTEXT:  PL/pgSQL function broken(uuid[],jsonb[]) line 5 at RETURN QUERY

>>> calling the 040 fix:
             broadcast_id             |             recipient_id             |              contact_id
--------------------------------------+--------------------------------------+--------------------------------------
 c6a2d459-1eb3-48bf-b98a-d44b312e3d34 | 8cb36b24-61cd-46ac-b874-7d8933cb4bae | 2586670c-632f-4a4c-b333-5cbfb1a5330e
(1 row)
```

Also confirmed against a live Supabase project inside a `BEGIN … ROLLBACK`: the function creates the campaign and returns its recipient row, nothing persisted.

## Two things worth a follow-up (not in this PR)

**A plpgsql body is only parsed at `CREATE` time — name resolution happens on first execution.** Applying a migration proves the function exists, not that it can run, so a check that applies migrations without *calling* the RPCs they define will go green on a function that is dead on arrival. That is exactly what happened here. A smoke test that invokes each RPC once would have caught this the day `037` landed.

**The blast radius hid it.** `broadcast-core.ts` is the only caller, reached from the public API. The dashboard's own broadcast route (`POST /api/whatsapp/broadcast`) loops `sendTemplateMessage` and never writes a campaign row, so the UI looks healthy while `broadcasts` and `broadcast_recipients` stay empty — which is why this survived from `037` (#370) through `038` (#472) unnoticed.

### DIFF
--- a/supabase/migrations/040_fix_broadcast_contact_id_ambiguity.sql
+++ b/supabase/migrations/040_fix_broadcast_contact_id_ambiguity.sql
@@ -1,0 +1,111 @@
+-- ============================================================
+-- 040_fix_broadcast_contact_id_ambiguity.sql — make
+--     create_broadcast_with_recipients executable
+--
+-- The problem
+--
+--   Every call to POST /api/v1/broadcasts dies in the database with
+--   SQLSTATE 42702:
+--
+--     column reference "contact_id" is ambiguous
+--     It could refer to either a PL/pgSQL variable or a table column.
+--
+--   `create_broadcast_with_recipients` is declared
+--   `RETURNS TABLE(broadcast_id UUID, recipient_id UUID, contact_id UUID)`,
+--   and in PL/pgSQL a RETURNS TABLE output column is ALSO an in-scope
+--   variable. The recipient INSERT ends in a bare
+--   `RETURNING id, contact_id`, so that `contact_id` resolves against
+--   both the target table's column and the function's own output
+--   variable, and Postgres refuses to guess. Qualifying it —
+--   `broadcast_recipients.contact_id` — names the column and nothing
+--   else. That one word is the entire fix.
+--
+--   The other identifiers in the body are already unambiguous:
+--   `broadcast_id` appears only in an INSERT column list (never a
+--   variable reference), and the final SELECT reads through `ins`.
+--
+-- Why nothing caught it
+--
+--   A plpgsql body is only parsed at CREATE time — name resolution
+--   happens on first EXECUTION. The migration applies cleanly, so both
+--   a fresh `supabase db reset` and CI go green on a function that
+--   cannot run. Worth considering a smoke test that CALLS the RPCs the
+--   migrations define, not just one that applies them.
+--
+--   The blast radius also hid it. `lib/whatsapp/broadcast-core.ts` is
+--   the only caller, reached from the public API. The dashboard's own
+--   broadcast route (POST /api/whatsapp/broadcast) loops
+--   sendTemplateMessage and never writes a campaign row, so the UI
+--   looks healthy while `broadcasts` and `broadcast_recipients` stay
+--   empty.
+--
+-- Introduced by 037 (which added the function, fixing #370) and
+-- carried forward unchanged by 038 (which added p_template_params,
+-- fixing #472). It has never once succeeded.
+--
+-- Why a new file rather than an edit to 038
+--
+--   Applied migrations are recorded in `schema_migrations`, so editing
+--   038 in place would fix only fresh installs — every existing
+--   deployment already has 038 recorded and would keep the broken
+--   function forever. Same reasoning, and the same shape, as
+--   034_fix_profiles_update_rls.sql repairing 017's policy.
+--
+--   This is a CREATE OR REPLACE of the exact 038 signature, so it is
+--   idempotent and safe to re-run. Signature, arguments and result
+--   columns are unchanged — broadcast-core.ts needs no edit.
+-- ============================================================
+
+CREATE OR REPLACE FUNCTION public.create_broadcast_with_recipients(
+  p_account_id        UUID,
+  p_user_id           UUID,
+  p_name              TEXT,
+  p_template_name     TEXT,
+  p_template_language TEXT,
+  p_total_recipients  INTEGER,
+  p_contact_ids       UUID[],
+  p_template_params   JSONB[]
+)
+RETURNS TABLE(broadcast_id UUID, recipient_id UUID, contact_id UUID)
+LANGUAGE plpgsql
+SECURITY DEFINER
+SET search_path = public
+AS $$
+DECLARE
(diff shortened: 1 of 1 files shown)


neerajgupta2407:
@ArnasDon Could you Please review this?

ArnasDon:
Merged via #566 with your commit and authorship intact — thank you for the precise diagnosis and the one-word fix. The only change was the file number: #533 landed as migration 040 while this was open, so yours ships as `041_fix_broadcast_contact_id_ambiguity.sql`. I also took your suggestion about execution-time name resolution and added a CI assertion on `pg_get_functiondef` so a replay that keeps the 038 body fails the migrations job. Closing this one since the branch can't be renumbered from here.

PR #545, opened 2026-09-03
feat(csat,leads): post-sale satisfaction survey + auto-cool stale leads

angelsandia-rgb:
Two evidence-layer features from the SANDÍA B2B plan — **step 1** (instrument satisfaction) and **step 2** (make lead temperature decay on its own). Most of "steps 1 & 2" already existed (temperature field, `/kpis` funnel, Google Sheets export); this fills the two real gaps.

## 1. CSAT post-venta

**Migration `102`** — `csat_config` (per-account: toggle, template, scale 3/5, delay, per-contact cooldown) + `csat_surveys` (one row per won deal, unique on `deal_id`).

- On **`deal.won`**, `dispatchCsat` — hooked *inside* `dispatchWebhookEvent`, exactly like `dispatchToGoogleSheets`, so all 4 deal.won sources are covered with zero new call sites — queues a survey row.
- **`GET /api/csat/cron`** (every 15 min) sends it once `send_after` elapses, via an approved **WhatsApp template with rating buttons** (`sendMessageToConversation`, `senderType: 'bot'`). `delay_minutes: 0` sends inline instead.
- The customer's **button tap is captured** in the inbound webhook path: `message.received` now carries `interactive_reply_id`; the Zernio WhatsApp path (which doesn't parse taps) accepts a bare `"5"` / `"★★★★★"` text *only while a survey is outstanding for that contact*. → status `responded` + new **`csat.received`** event that fans out through the same path (webhooks + a new **"Satisfacción"** Google Sheets tab).
- **Settings → Satisfacción (CSAT)** config card (new rail section). **Average-score `MetricCard` on `/kpis`** (normalised by scale so 3- and 5-button accounts compare).

## 2. Lead-temperature auto-cool

**Migration `103`** — `contacts.lead_temperature_updated_at` (backfilled) + `accounts.lead_cooldown_enabled` / `lead_cooldown_days`.

- Opt-in per account in **Settings → Deals & currency**.
- **`GET /api/contacts/temperature-sweep/cron`** (hourly) steps warm/hot contacts one notch cooler (`hot→warm→cold`) once the thread has been silent **and** the temperature has held for `lead_cooldown_days` (a fresh classification gets its own grace period). Each change emits `contact.lead_temperature_changed`, same as the manual inbox control.
- `lead_temperature_updated_at` is now also stamped by the manual PATCH and both AI set-temperature paths.

## Migration `104` + ops

Registers both pg_cron jobs (psql-var style like `100`; reuse `AUTOMATION_CRON_SECRET`). Heartbeats `csat_cron` + `temperature_sweep_cron` added to the registry — they read "never" (warning alert) until the jobs are registered, same as `followups_cron`.

**To deploy:** apply migrations 102 & 103, then run 104 (or add the two `cron.schedule` calls) with the base URL + secret filled in.

## Checks

`npm run lint`, `npm run typecheck`, `npm test` (1348 passed, +37 new), `npm run build` — all green. New unit tests: `src/lib/csat/config.test.ts`, `src/lib/contacts/temperature-cooldown.test.ts`.


angelsandia-rgb:
Opened against the wrong repo (upstream template). Reopening on angelsandia-rgb/wacrm.

PR #554, opened 2026-09-06
fix(whatsapp): keep the stored verify_token when config is saved without one

Guru35:
## Summary

Saving the WhatsApp settings form clears `whatsapp_config.verify_token` unless
the token field happens to be filled in — which it never is, because the form
does not render the stored value. Every save of any *other* field (phone
number id, WABA id, access token…) silently nulls the verify token, and Meta's
webhook GET handshake then returns **403**. From the outside this is
indistinguishable from "you typed the wrong token".

Root cause is one expression in the save path:

```ts
verify_token: incoming ? encrypt(incoming) : null
```

An empty field means *"keep what is stored"*, not *"clear it"* — the settings
GET endpoint deliberately never returns the token (it is encrypted at rest),
so empty is the normal state of that input.

Measured on a live self-hosted instance (2026-09-05): the row held
`verify_token = NULL`, Meta's verification GET returned 403, inbound message
delivery stopped. Restoring the token fixed it; saving any other field broke
it again.

## What changed

* `src/lib/whatsapp/verify-token.ts` — new `resolveVerifyTokenForSave(incoming, existingEncrypted)`:
  a non-empty value is encrypted and stored, an empty/whitespace value keeps
  the existing ciphertext. Clearing a token on its own is not a supported
  operation; **Reset Configuration** already wipes the whole row.
* `src/app/api/whatsapp/config/route.ts` — save path reads the existing row
  and delegates to the helper instead of inlining the ternary.
* `src/components/settings/whatsapp-config.tsx` — the field now says the token
  is kept when left blank, so the behaviour is discoverable rather than
  implied.
* `messages/en.json`, `messages/ko.json` — that one hint string.

Behaviour is unchanged when a token *is* typed.

## Test plan

* [x] `src/lib/whatsapp/verify-token.test.ts` — **7 passed**, written
  red-first (the test commit precedes the fix commit in this PR).
* [x] **Negative control:** replacing `return existingEncrypted ?? null` with
  `return null` (the old behaviour) turns **4 of the 7 red**; reverting makes
  them green again. The suite genuinely pins this behaviour rather than
  passing vacuously.
* [x] `npm run typecheck` — clean.
* [x] `npx eslint` on the four touched files — **0 errors, 1 warning**, and
  that warning (`react-hooks/exhaustive-deps` in `whatsapp-config.tsx`) is
  **pre-existing**: it reports on `upstream/main` too, at line 204 vs 210 here
  — the shift is just the added lines.
* [ ] `npm run build` — not run locally; left to CI.

## Related

Found while self-hosting; no existing issue covers it (searched open + closed
for `verify_token`). Happy to split the i18n hint into its own commit or drop
it if you would rather keep the fix minimal.

### DIFF
--- a/messages/en.json
+++ b/messages/en.json
@@ -1497,6 +1497,7 @@
       "webhookVerifyToken": "Webhook Verify Token",
       "webhookVerifyTokenPlaceholder": "Create a custom verify token",
       "webhookVerifyTokenHint": "A custom string you create. Must match the token you set in Meta webhook settings.",
+      "webhookVerifyTokenSaved": "Saved. Leave as is to keep it, or enter a new value to rotate.",
       "twoStepPin": "Two-step verification PIN",
       "optional": "(optional)",
       "pinPlaceholder": "6-digit PIN from Meta WhatsApp Manager",
--- a/messages/ko.json
+++ b/messages/ko.json
@@ -1497,6 +1497,7 @@
       "webhookVerifyToken": "웹훅 인증 토큰",
       "webhookVerifyTokenPlaceholder": "사용자 지정 인증 토큰을 만드세요",
       "webhookVerifyTokenHint": "직접 만드는 문자열입니다. Meta 웹훅 설정에 지정한 토큰과 일치해야 합니다.",
+      "webhookVerifyTokenSaved": "저장되어 있습니다. 유지하려면 그대로 두고, 교체하려면 새 값을 입력하세요.",
       "twoStepPin": "2단계 인증 PIN",
       "optional": "(선택)",
       "pinPlaceholder": "Meta WhatsApp 관리자에서 발급받은 6자리 PIN",
--- a/src/app/api/whatsapp/config/route.ts
+++ b/src/app/api/whatsapp/config/route.ts
@@ -7,6 +7,7 @@ import {
   verifyPhoneNumber,
 } from '@/lib/whatsapp/meta-api'
 import { encrypt, decrypt } from '@/lib/whatsapp/encryption'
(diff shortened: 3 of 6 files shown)


ArnasDon (changes requested):
Thanks — the server-side fix is right and the red-first test pins it well. I merged this onto current `main` (which now includes #578, so the `whatsapp-config.tsx` rebase is already accounted for): merges cleanly, `eslint` 0 errors, `tsc --noEmit` clean, vitest 999/999, `next build` passes.

One thing needs fixing before merge — the UI half of the change is inert:

- `GET /api/whatsapp/config` selects only `phone_number_id, waba_id, access_token, status` (`src/app/api/whatsapp/config/route.ts`, `GET`), so `data.verify_token` is always `undefined` on the client. That means `setVerifyToken(data.verify_token ? MASKED_TOKEN : '')` always sets `''`, and `config?.verify_token && !verifyEdited` is always false, so the new mask and the "Saved. Leave as is to keep it…" hint never render. The `webhookVerifyTokenSaved` key (en + ko) is dead.
- Suggested fix: have `GET` return a boolean such as `has_verify_token: !!row.verify_token` (never the ciphertext), add it to the `WhatsAppConfig` type, and key the mask + hint on that. Alternatively drop the mask/hint and keep just the server change plus the `verifyEdited` payload gating — but the discoverability point in your description is a good one, so I'd go with the boolean.

Minor: with the mask cleared `onFocus`, blurring without typing leaves an empty field with the generic hint even though the token is still stored. Keying the hint on the server boolean rather than `verifyEdited` fixes that too.

Happy to merge once the GET side is wired up.

Issue #560, opened 2026-09-11
Public API contact tag replacement returns 200 but leaves unrelated tags


itsbkm:
## Summary

`PATCH /api/v1/contacts/{id}` returns HTTP 200 but does not replace the contact's tags as documented.

## Environment

- WACRM upstream: `ArnasDon/wacrm`
- Deployed commit: `98b5bd2`
- Self-hosted Next.js deployment with Supabase

## Reproduction

For contact `ae9f26a4-3d05-4b77-b1ff-39b81afb9cf1` (phone `6005010475`), send:

```http
PATCH /api/v1/contacts/ae9f26a4-3d05-4b77-b1ff-39b81afb9cf1
Content-Type: application/json
Authorization: Bearer <api-key>

{"tags":["ipad","3y","jmu"]}
```

The request returns HTTP 200. A fresh GET of the same contact still returns the previous tag set, including unrelated tags such as `hsp`.

## Observed persisted state

Directly querying Supabase after the PATCH shows:

- 15 `contact_tags` rows remain for the contact
- tags include `1y, 2y, 3y, 4y, 5y+, chd, hsp, hsr, ipad, iphone, jmu, kng, mac, moh, zrk`
- `contacts.updated_at` remains equal to `created_at`

The unchanged contact timestamp may be expected if tag writes do not update the parent contact row, but the unchanged `contact_tags` rows are not expected.

There are no active `tag_added` automations in the account, and no database trigger was found that attaches all account tags to a contact.

## Expected behavior

The documented API says passing `tags` replaces the contact's tags. After the request, only `ipad`, `3y`, and `jmu` should remain.

The same over-tagging was observed during initial contact creation: contacts acquired the account's growing tag catalog instead of only the tags in their individual payloads.

## Documentation

The public API documentation states that `PATCH /api/v1/contacts/{id}` with `tags` replaces the contact's tags: https://github.com/ArnasDon/wacrm/blob/main/docs/public-api.md

Please advise whether this is a known persistence issue or whether there is another required request shape/endpoint for exact per-contact tag replacement.

itsbkm:
## Clean reproduction after deleting all contacts

This reproduces the create-path issue with no possible accumulation or retry effects.

- Account had **0 contacts** before the request.
- Account had only the 15 pre-existing tag definitions.
- Exactly one request was sent; the full request/response is captured in the integrator log.

Request:

```json
{
  "phone": "+6000486114",
  "tags": ["5y+", "moh"]
}
[...]

PR #561, opened 2026-09-11
fix(api): verify contact tag replacement persists

itsbkm:
## Problem

The public contact API can return HTTP 200 after a tag replacement request even when stale `contact_tags` rows remain. Tag-only updates also leave `contacts.updated_at` unchanged, making successful changes difficult to observe.

Issue: #560

## Fix

- Request the deleted tag IDs back from PostgREST and fail if any stale join was not removed.
- Update the parent contact's `updated_at` when the tag set changes.
- Keep the existing exact replacement semantics: the supplied tag names are the complete desired set.

This prevents a silent success response when persistence did not match the requested tag set.

## Validation

- `git diff --check` passes.
- The code path is account-scoped and continues to use the existing service-role API client.
- Full TypeScript validation should run in CI; the checkout does not currently have `node_modules` installed.


itsbkm:
Validation: `docker compose --env-file .env.local build app` passed, including the Next.js production build and TypeScript check. `git diff --check` also passes.

itsbkm:
New clean reproduction added to issue #560: with 0 contacts and one POST payload {"tags":["5y+","moh"]}, the created contact receives all 15 account tags. This confirms the create path has a separate over-tagging defect beyond the PATCH persistence symptom.

ArnasDon:
Thanks for the report and for digging into this — #560 was real, but the cause turned out to be one level up from where this PR looks. `resolveImportTagIds` returns a map of **every** tag in the account (it loads them all for case-insensitive matching), and `setContactTags` was taking `tagIdByKey.values()` wholesale as the desired set. So the delete step wasn't failing silently; it was never asked to remove anything, and the add step attached the whole catalog. That's why a fresh account with 15 tag definitions produced a 15-tag contact.

Fixed in #564 (merged) by deriving the desired set from the requested names only, with a regression test that fails on `main`. The verification added here would still have returned 200 on that path, so I'm closing this in favour of #564 rather than layering it on top. Appreciate the clean repro in the issue — it made the cause unambiguous.

